PR #3, opened 2016-05-09
Fix typo in migration example

davydovanton:


jeremyevans:
The `db` variable is necessary here.  `create_table` uses instance_eval, so `database_type` calls a different method with your change.

The change to `deadline_opts` in the migration is also wrong, as `remember_deadline_interval` is documented as being 14 days by default, not 1.

Note that the migration in the README is directly copied out of the `spec/migrate/001_tables.rb` file, so if these were problems, the tests would break.

Did you actually have problems running the migration in the README?  If so, can you post your environment and any backtrace?

davydovanton:
hm, thank you for explanation. I think I understand 👍 
No, I haven't any problems but I copied this migration and found strange places of code which I wanted to improve.

PR #8, opened 2016-09-29
Add support for JWT claims and verifiers

mwpastore:
* Add jwt_foo and jwt_verify_foo? auth value methods to jwt feature and pass these along to ruby-jwt on encode/decode
* Nest session data under 'data' key in payload to protect against collisions with reserved claims (BREAKING CHANGE)
* Deeply symbolize session hash after decoding jwt payload (POTENTIALLY BREAKING CHANGE)

### DIFF
--- a/lib/rodauth/features/jwt.rb
+++ b/lib/rodauth/features/jwt.rb
@@ -26,6 +26,20 @@ module Rodauth
       :use_jwt?
     )
 
+    # JWT claims and verifiers (see ruby-jwt for more info)
+    auth_value_method :jwt_sub, proc { |ra| ra.session[ra.session_key] }
+    auth_value_method :jwt_verify_sub?, false
+    auth_value_method :jwt_iat, proc { Time.now.to_i }
+    auth_value_method :jwt_verify_iat?, true
+    auth_value_method :jwt_jti, nil
+    auth_value_method :jwt_verify_jti?, false
+    auth_value_method :jwt_leeway, 30
+
+    %i[aud exp iss nbf].each do |reserved_claim|
+      auth_value_method "jwt_#{reserved_claim}".to_sym, nil
+      auth_value_method "jwt_verify_#{reserved_claim}?".to_sym, true
+    end
+
     auth_methods(
       :json_request?,
       :jwt_token,
@@ -37,12 +51,8 @@ module Rodauth
       return @session if defined?(@session)
       return super unless use_jwt?
 
-      @session = if jwt_token
-        s = {}
-        jwt_payload.each do |k,v|
-          s[k.to_sym] = v
-        end
-        s
+      @session = if jwt_token && data = jwt_payload['data']
+        JSON.parse(JSON.fast_generate(data), :symbolize_names=>true)
       else
         {}
       end
@@ -83,7 +93,55 @@ module Rodauth
     end
 
     def session_jwt
-      JWT.encode(session, jwt_secret, jwt_algorithm)
+      JWT.encode(jwt_claims.merge(:data=>session), jwt_secret, jwt_algorithm)
+    end
+
+    def jwt_claims
+      claims = {}
+      %i[aud exp iat iss jti nbf sub].each do |reserved_claim|
+        next unless claim = send("jwt_#{reserved_claim}")
+
+        claims[reserved_claim] =
+          if claim.respond_to?(:call)
+            claim.call(self)
+          else
+            claim
+          end
+      end
+      claims
+    end
+
+    def jwt_verifiers
+      verifiers = {}
+      %i[aud exp iat iss jti nbf sub].each do |reserved_claim|
+        # if there is no claim to verify or jwt_verify_foo? is falsy, we have nothing to do
+        next unless send("jwt_#{reserved_claim}") && verifier = send("jwt_verify_#{reserved_claim}?")
+
+        verifiers["verify_#{reserved_claim}".to_sym] = true
+
+        # time-based verifiers do not require a comparison operand
+        next if %i[exp iat nbf].include?(reserved_claim)
+
+        if reserved_claim == :jti
+          if verifier.respond_to?(:call) && (verifier = verifier.call(self)).respond_to?(:call)
+            verifiers[reserved_claim] = verifier
+            next
+          end
+
+          raise ArgumentError, 'JWT ID verifier must be callable (and return a callable)'
+        end
+
+        verifiers[reserved_claim] =
+          if verifier.respond_to?(:call)
+            verifier.call(self)
+          elsif verifier == true
+            send("jwt_#{reserved_claim}", self)
+          else
+            verifier
+          end
+      end
+      verifiers[:leeway] = jwt_leeway if jwt_leeway
+      verifiers
     end
 
     def jwt_token
@@ -134,7 +192,7 @@ module Rodauth
     end
 
     def jwt_payload
-      JWT.decode(jwt_token, jwt_secret, true, :algorithm=>jwt_algorithm)[0]
+      JWT.decode(jwt_token, jwt_secret, true, jwt_verifiers.merge(:algorithm=>jwt_algorithm))[0]
     rescue JWT::DecodeError
       json_response[json_response_error_key] = invalid_jwt_format_error_message
       response.status ||= json_response_error_status
--- a/spec/jwt_spec.rb
+++ b/spec/jwt_spec.rb
@@ -122,4 +122,34 @@ describe 'Rodauth login feature' do
     json_request("/login", :headers=>{'HTTP_ACCEPT'=>'application/*'}, :login=>'foo@example.com', :password=>'0123456789').must_equal [200, {"success"=>'You have been logged in'}]
     json_request("/login", :headers=>{'HTTP_ACCEPT'=>'application/vnd.api+json'}, :login=>'foo@example.com', :password=>'0123456789').must_equal [200, {"success"=>'You have been logged in'}]
   end
+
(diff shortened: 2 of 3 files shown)


lib/rodauth/features/jwt.rb
```diff
@@ -37,12 +51,8 @@ def session
...
       return super unless use_jwt?
 
-      @session = if jwt_token
-        s = {}
-        jwt_payload.each do |k,v|
-          s[k.to_sym] = v
-        end
-        s
+      @session = if jwt_token && data = jwt_payload['data']
+        JSON.parse(JSON.fast_generate(data), :symbolize_names=>true)
```
mwpastore: Replacing a shallow symbolization with a deep one. Unfortunately, ruby-jwt doesn't let you "punch down" the `:symbolize_names` option on decode. I think we can get away with `.fast_generate` here because we know the data was recently acceptable JSON, i.e. no cycles.

lib/rodauth/features/jwt.rb
```diff
@@ -26,6 +26,20 @@ module Rodauth
...
     )
 
+    # JWT claims and verifiers (see ruby-jwt for more info)
+    auth_value_method :jwt_sub, proc { |ra| ra.session[ra.session_key] }
+    auth_value_method :jwt_verify_sub?, false
+    auth_value_method :jwt_iat, proc { Time.now.to_i }
+    auth_value_method :jwt_verify_iat?, true
+    auth_value_method :jwt_jti, nil
+    auth_value_method :jwt_verify_jti?, false
+    auth_value_method :jwt_leeway, 30
```
mwpastore: I think this is an OK default, to account for some small amount of clock skew on a small cluster of nodes serving a JSON API, but perhaps it would be better to neither include nor verify an iat claim by default.

lib/rodauth/features/jwt.rb
```diff
@@ -83,7 +93,55 @@ def set_notice_now_flash(message)
...
+        # time-based verifiers do not require a comparison operand
+        next if %i[exp iat nbf].include?(reserved_claim)
+
+        if reserved_claim == :jti
+          if verifier.respond_to?(:call) && (verifier = verifier.call(self)).respond_to?(:call)
+            verifiers[reserved_claim] = verifier
+            next
+          end
+
+          raise ArgumentError, 'JWT ID verifier must be callable (and return a callable)'
```
mwpastore: Is this the correct type of exception to throw?

lib/rodauth/features/jwt.rb
```diff
@@ -83,7 +93,55 @@ def set_notice_now_flash(message)
...
+    end
+
+    def jwt_claims
+      claims = {}
+      %i[aud exp iat iss jti nbf sub].each do |reserved_claim|
+        next unless claim = send("jwt_#{reserved_claim}")
+
+        claims[reserved_claim] =
+          if claim.respond_to?(:call)
+            claim.call(self)
```
mwpastore: Is passing the Rodauth instance to the proc the best way for the caller to introspect the application to generate a jti/sub/aud/etc. claim? The one big caveat is that you can't access the session when verifying the session (it recurses infinitely), so we either need a guard against that or some kind of warning in the documentation.

spec/jwt_spec.rb
```diff
@@ -122,4 +122,34 @@
...
+    payload, _header = JWT.decode @authorization, nil, false
+    payload['sub'].must_equal payload['data']['account_id']
+    payload['iat'].must_be_kind_of Integer
+    payload['exp'].must_be_kind_of Integer
+    payload['nbf'].must_be_kind_of Integer
+    payload['iss'].must_equal "Foobar, Inc."
+    payload['aud'].must_equal %w[Young Old]
+    payload['jti'].must_match(/^[0-9a-f]{8}-?[0-9a-f]{4}-?[0-5][0-9a-f]{3}-?[089ab][0-9a-f]{3}-?[0-9a-f]{12}$/i)
+
+    json_request.must_equal [200, [1]]
```
mwpastore: I feel like these tests are insufficient but short of manually encoding bad JWTs to send in and fail verification, I don't know what else to do.

jeremyevans:
First, thanks for submitting a pull request.

Can you explain why this is needed?  Rodauth is only designed to consume JWTs that it itself generates, it is not designed to be used with JWTs generated from external applications, and it is not designed to produce JWTs to be consumed by other applications. Basically, JWT is simply used as an internal data storage format, and not for integration with external systems. As Rodauth sets no reserved claims, I don't see why it would need to put all claims it does set under a key, and I don't know why we would need these additional jwt_* methods.  As Rodauth does not set claims with subhashes, I don't see why a deep symbolization of keys is necessary.

mwpastore:
Sure. Setting aside implementation details for a moment, let's talk about what I'm trying to do. In most applications, JWTs replace not only the session cookie but also the session. This removes the shared state between the service's backend servers and is a key component of a RESTful service. My goal is to use JWTs as I've used them in past applications but allow Rodauth to generate and consume them (while also using Rodauth's many other features); and to store session data in the JWT payload, accessing it through `env['rodauth'].session` on the backend. I will not be using a session store based on Rack::Session::Abstract::Persisted or anything else as it would be completely superfluous.

I'd like to be able to store different types of data in my session: arrays, hashes, etc., and since it's just being serialized to JSON and stored in the JWT payload, this should be perfectly valid:

```ruby
env['rodauth'].session[:info] = { :foo => [0, 1, 2], :bar => { :qux => "frob" } }
```

Later, when I'm inspecting the session of an existing JWT payload, I want it to look like the above, not like:

```ruby
env['rodauth'].session[:info] # => { 'foo' => [0, 1, 2], 'bar' => { 'qux' => "frob" } }
```

So after decoding the JWT payload, I'd like any subhashes to be deeply symbolized.

Because I don't have the concept of a traditional session, nor a cookie, I now need a way to specify the lifetime of my JWT. Let's say I don't want my JWTs to be valid for longer than 30 days. That's traditionally done by setting the [`exp`](https://github.com/jwt/ruby-jwt#expiration-time-claim) reserved claim on the JWT. In the current implementation of Rodauth, I could potentially do this:

```ruby
rodauth.session[:exp] ||= Time.now.to_i + 30 * 24 * 60 * 60
rodauth.session[:verify_exp] = true
```

Other than the weirdness of doing this in an `after_login` block or something and generating the JWT a few extra times, I suppose this gets the job done. The other catch would be making sure I or someone else on my team don't accidentally use the `:exp` element of my session hash for something else. This is probably a bad example because Rodauth has the session_expiration feature, which works almost the *exact same way* as the `exp` reserved claim. (As an aside, it might be interesting to modify how the session_expiration feature works when the jwt plugin is loaded to use the `exp` claim instead of its own logic.)

Here's another example: let's say I assign each JWT a unique ID and keep track of which JWT IDs belong to which users. I also maintain an eventually-consistent lookup table of blacklisted JWT IDs on each server. When a user logs out or reports suspicious activity on their account, I add the appropriate JWT IDs to the backlist. JWT has a mechanism for this in the [`jti`](https://github.com/jwt/ruby-jwt#jwt-id-claim) reserved claim:

```ruby
rodauth.session[:jti] ||= SecureRandom.uuid.tap do |uuid|
  associate_jti_with_user(uuid, rodauth.session[:account_key])
end
rodauth.session[:verify_jti] = proc { |jti| !blacklisted_jti?(jti) }
```

This is kind of weird because a proc can't be serialized to JSON, and we still have all the same problems as with the `exp` reserved claim. 

Assuming you're with me so far and think it might be an okay idea to add support for JWT reserved claims to the jwt feature of Rodauth, let's talk about implementation. Hopefully you see that keeping these reserved claims and verification scalars and procs at the same level as the rest of the session data is probably a bad idea. We could put all the reserved claims and verifiers in a big hash and let the caller pass in a different hash. I tried to follow the examples of the existing features in that you prefer different auth value methods for related things, e.g. `:jwt_authorization_ignore` and `:jwt_authorization_remove` c.f. a single `:jwt_authorization` hash with `:ignore` and `:remove` keys. I can take a completely different approach here if you see a better way to do it. 

Thanks Jeremy!

jeremyevans:
Thank you for your detailed description of your situation and why this pull request would be beneficial for it.

For automatically expiring older sessions, as you mentioned, instead of using the JWT `exp` claim, you can use the Rodauth `session_expiration` feature.  I think that's a better approach, as it is more generic and not specific to JWT.

For the other claims, they seem to be fairly specific to your use case, and I don't think the use case is common enough to warrant specific support for them in Rodauth, yet at least.  It's possible that's due to my lack of experience with JWT, so maybe more experience will change my mind in the future.  At present, I just want to use JWT as a data storage format, without using any of the reserved claims.

I would definitely support refactoring in the jwt feature such that you could easily add your own claims before creating the jwt, and verify the claims after, if the current configuration methods are not sufficient for that purpose.  We could then document how to use those configuration methods to add support for the reserved claims.

For putting all session data under a separate key in the JWT, I see the problem with the current approach, and am OK with an option to do that as long as it is off by default for backwards compatibility (we can turn it on by default in Rodauth 2.0).

I'm also fine with an option for deep symbolization of the session hash keys, again as long as it is off by default (and it would probably stay off by default in Rodauth 2.0).

Does the approach I've described seem reasonable?

mwpastore:
You're welcome. I hope I didn't come across as condescending, I just wanted to make sure I covered all the bases.

Yes, that makes sense. Let me take a stab at it and update this PR (or create a new one).

jeremyevans:
Your post and comments were definitely not condescending, I learned quite a bit from them.  I look forward to reviewing your future commits, be they in this PR or a separate one.

PR #9, opened 2016-09-30
Add basic support for JWT claims and verifiers

mwpastore:
* Add jwt_claims and jwt_verifiers auth value methods to jwt feature and pass these along to ruby-jwt on encode/decode.
* Add jwt_nest_session? auth value method to optionally nest the session hash in the JWT payload under the 'data' key.
* Add jwt_symbolize_deeply? auth value method to optionally symbolize subhash keys when decoding the JWT payload.

Follow-up to #8!


jeremyevans:
Thanks for working on this.  These changes look good, I'll run them through the tests now.

mwpastore:
Sure thing! Feel free to leave line comments and I'll make whatever changes you'd like. Thanks!

jeremyevans:
After running some tests and taking a closer look at the API, I ended up significantly refactoring this pull request after merging.  Please take a look at the revised API  (e1c258ea707cddbcb391399ef540ccfd1f85414b) and let me know if you think we should make further changes.

PR #10, opened 2016-09-30
Memoize jwt_payload method

mwpastore:
File under "shooting myself in the foot": now that the session data is nested, in order to retrieve reserved claims from the JWT you need to write e.g. `jwt_payload['jti']`. This of course incurs an expensive call to JWT.decode each time. Here's a quick and dirty fix.


jeremyevans:
That certainly looks reasonable to me, I'll merge and test now.

PR #16, opened 2016-10-15
Fix basic auth

### LINKED ISSUE #13: HTTP Basic auth failing with curl
[The regexp is somehow not parsing out the baisc token when coming from cURL](https://github.com/jeremyevans/rodauth/blob/master/lib/rodauth/features/http_basic_auth.rb#L11).

* First, it seems that the newline (which I added because it was appended on the rack test helpers) may or may not be there (curl doesn't). 
* That messes up the `\z` (if removed, it all works). Do you have any idea why this happens?

TiagoCardoso1983:
To fix #13 

I said I was going with splitting, but in order to keep the same multi-conditionals you rewrote my previous commits into, I could "split then assign then compare to scheme". This seems the least-effort fix. 

However, I leave you with a small script showing the differences between the 2 approaches, and you can later decide whether to switch to string splitting and how. 

```ruby
require 'benchmark/ips'
 Benchmark.ips do |x|
   HEADER = "Basic 2347z8h4r378043ef8ju43ew==\n"
 
   x.report("splitting") { HEADER.split(' ', 2)[1]  }
   x.report("matching") { HEADER[/\A *Basic (.*)\Z/, 1] }
 
   x.compare!
 end
```

```
Warming up --------------------------------------
           splitting    90.472k i/100ms
            matching    47.837k i/100ms
Calculating -------------------------------------
           splitting      1.500M (± 4.9%) i/s -      7.509M in   5.019272s
            matching    805.981k (± 5.1%) i/s -      4.066M in   5.058974s

Comparison:
           splitting:  1500012.1 i/s
            matching:   805980.9 i/s - 1.86x  slower
```

### DIFF
--- a/lib/rodauth/features/http_basic_auth.rb
+++ b/lib/rodauth/features/http_basic_auth.rb
@@ -8,7 +8,7 @@ module Rodauth
       return @session if defined?(@session)
       sess = super
       return sess if sess[session_key]
-      return sess unless token = ((v = request.env['HTTP_AUTHORIZATION']) && v[/\A *Basic (.*)\n\z/, 1])
+      return sess unless token = ((v = request.env['HTTP_AUTHORIZATION']) && v[/\A *Basic (.*)\Z/, 1])
       username, password = token.unpack("m*").first.split(/:/, 2)
 
       if username && password
--- a/lib/rodauth/features/http_basic_auth.rb
+++ b/lib/rodauth/features/http_basic_auth.rb
@@ -1,0 +1,7 @@
+# frozen-string-literal: true
+
+module Rodauth
+  HTTTBasicAuth = Feature.define(:http_basic_auth) do
+
+  end
+end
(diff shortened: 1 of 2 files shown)


jeremyevans:
I think the patch here is fine.  The benchmark is not really applicable as it doesn't include a check that the first element is "Basic".  Even if the difference is about 2x, I think the approach in this patch is a bit cleaner, and this code is unlikely to be a bottleneck.

We should probably add a spec that checks both with and without the `\n` to make sure things work in both cases.

jeremyevans:
Spec looks good.  Could you squash rebase against master?  If not, I'll take care of that when I merge this tomorrow.

jeremyevans:
I squash merged this at 4dc752f73fb50c71abb199b4417f6681aedb13fc

TiagoCardoso1983:
@jeremyevans sorry, didn't find the time to do it earlier. Thx again.

PR #21, opened 2017-02-13
Add external session login example.

ayghor:
it is still a wip, but if you say it is cool enough, i will improve. wat do u think? <3


jeremyevans:
Thanks.  I think this is a good start.  I'll try to merge it tomorrow and try to make it a bit more generic.

PR #25, opened 2017-02-20
Suggest recovery codes be securely stored out-of-band

adam12:
Make it incredibly obvious how these tokens should be used.


jeremyevans:
This looks good, thanks for the patch!

PR #28, opened 2017-02-28
Implementing short access token, with a longer lived DB-stored refresh token

allavena:
I needed a short access token, with longer lived refresh token.

Refresh tokens are stored in the DB and expired on use. They are provided in the JSON body of the response on login and calls to refresh. 

This allows to run Rodauth as a standalone Authorization server, with a separate API system simply checking the validity of the short-lived access tokens.

I made one change to the existing JWT Feature. I do not think a token should be provided when the login is rejected (even if that token did not contain the account).

Tests pass on Ruby 1.8.7 to 2.4. They seem to fail for the same reason as master on jruby.

### DIFF
--- a/README.rdoc
+++ b/README.rdoc
@@ -39,6 +39,7 @@ hashes by protecting access via database functions.
 * Session Expiration
 * Single Session (Only one active session per account)
 * JWT (JSON API support for all other features)
+* JWT Refresh (Access & Refresh Token)
 * Update Password Hash (when hash cost changes)
 * HTTP Basic Auth
 
@@ -269,6 +270,14 @@ versions of Sequel, switch the :Bignum symbols to Bignum constants.
       # Used by the password reset feature
       create_table(:account_password_reset_keys) do
         foreign_key :id, :accounts, :primary_key=>true, :type=>:Bignum
+        String :key, :null=>false
+        DateTime :deadline, deadline_opts[1]
+      end
+
+      # Used by the refresh token feature
+      create_table(:account_refresh_tokens) do
+        primary_key :id, :type=>:Bignum
+        foreign_key :account_id, :accounts, :type=>:Bignum
         String :key, :null=>false
         DateTime :deadline, deadline_opts[1]
       end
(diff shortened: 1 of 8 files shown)


jeremyevans:
Thank you for working on this.  However, it appears to me to be too custom to your specific needs, and I'm not sure if other Rodauth users would have the same needs as you.  You could publish as as an external gem (e.g. `rodauth-jwt_refresh`).  You could also ask on the Google Group and see if other Rodauth users would like it added.

If you could submit the JWT change as a separate pull request, I would appreciate it. I agree with you that we probably shouldn't return a JWT token in the error case.

The jruby failure on master should be fixed tomorrow when the new version of Sequel is released.

xvonabur:
Hello @jeremyevans ! I would like to have this feature implemented too. It's common nowadays to use jwt as combination of `refresh` and `access` tokens. For example, check `jwt_sessions` gem.

I think the best way is just create a new feature with a new name and put business logic from PR there. But I rather extend current `jwt` feature and change only necessary part. What's the best way to do that?

jeremyevans:
@xvonabur Were you able to use this pull request to implement what you need?  If so, maybe I can reopen it and merge it.

xvonabur:
@jeremyevans Yes, this PR covers all I need in my project.

jeremyevans:
@xvonabur Thanks for the feedback.  I'll spend more time reviewing the code and see if I am comfortable maintaining it and make sure it passes all of the tests on all supported databases.  If so, I'll merge it and it will be part of the next release, which should happen by mid-June.

jeremyevans:
I spent some time reviewing this today and making substantial changes to it (such as adding support for `hmac_secret`).  I'll do some more testing tomorrow, and assuming no problems the changes should be pushed to GitHub tomorrow.

xvonabur:
@jeremyevans That's great! I've temporary copied code from this PR to my project, but looking forward to delete it and use your version. Thanks for the effort.

jeremyevans:
This PR has been rebased, refactored, and merged into master: https://github.com/jeremyevans/rodauth/commit/1279f31ee7fc0cfda9385fefe0a7fa7fe794ab7d

Please try it out and let me know if you think there should be any changes before the next release (which should be in about a week).

PR #29, opened 2017-03-01
Don't provide a token when login fails

allavena:
This change removes providing a JWT when the user attempts to sign up, and is unsuccessful.

Open question, when clearing the session, the JWT is set as well. Is this correct behaviour?

### DIFF
--- a/lib/rodauth/features/jwt.rb
+++ b/lib/rodauth/features/jwt.rb
@@ -179,7 +179,7 @@ module Rodauth
 
     def return_json_response
       response.status ||= json_response_error_status if json_response[json_response_error_key]
-      set_jwt
+      set_jwt unless json_response[json_response_error_key]
       response['Content-Type'] ||= json_response_content_type
       response.write(request.send(:convert_to_json, json_response))
       request.halt


jeremyevans:
Thanks for working on this.  I'll try to test and merge it tomorrow.

When clearing the session, it is expected that the JWT will be set to an empty one, because clearing a session empties the session out, so the returned JWT will have an empty session.  The expectation is that every time a user sends a JWT, it will use a returned JWT if there is one as the JWT to use for future requests.

allavena:
Ahh.. There are no tests here, this was tested indirectly by my feature. It's a one liner though.  I may not get to it for a few days.

jeremyevans:
I tried to add a spec for this, and it fails.  The reason it fails is that `clear_session` sets the jwt early, before `return_json_response` is called.  So any login failures are automatically going to have a new jwt generated.  After some more thought, this matches the behavior of the browser based apps, so it's probably a good idea to keep the existing behavior.

If you really want to override the behavior in your custom feature, try this:

```ruby
  def set_jwt
    if json_response[json_response_error_key]
      response.headers.delete('Authorization')
    else
      super
    end
  end
```

PR #32, opened 2017-03-14
Update README.rdoc

1gor:
add drop_table  :account_login_change_keys to down migration


jeremyevans:
Good catch, thanks for the patch!

PR #34, opened 2017-04-18
Move update_session method to public visibility

adam12:


jeremyevans:

PR #39, opened 2017-08-03
Fix tabs in Readme

zishe:


jeremyevans:
Those aren't tabs, but it is one space too many. Thanks for the patch!

PR #40, opened 2017-09-14
Update index.erb

araslanov-e:


jeremyevans:
Can you please explain reason behind the change?  The previous code is valid, as this is a `config.ru` file and that supports use at top level.

araslanov-e:
Oy, sorry.

PR #42, opened 2017-09-18
Added a value method to silence an error when no matching login is found

kgorin:
This is needed to deny email probing attempts.


kgorin:
Sorry, wrong base.

PR #43, opened 2017-10-03
Have password reset flow respect deadlines.

chanks:
I noticed that the password reset flow wasn't actually respecting the deadline column. I didn't want to break any internal APIs that someone else might be relying on, so I made the simplest change I could think of, but of course feel free to tweak this however you like. Also I opted for the path of least transparency (making the output from a missing key identical to that of an expired key), but it might be preferable for this case to trigger a "that link has expired" message.

I also ran into a second issue with password reset, for which the fix is less clear to me: if the accounts primary key is when using an integer primary key and someone visits a password reset link like `/reset-password?key=blah_blah`, Rodauth doesn't recognize that the user id isn't valid and you get a `Sequel::DatabaseError: PG::InvalidTextRepresentation: ERROR:  invalid input syntax for integer: "blah"` 500 error. I don't think it's a security issue or anything, but I guess someone spamming tokens would result in a lot of error reports.

### DIFF
--- a/lib/rodauth/features/reset_password.rb
+++ b/lib/rodauth/features/reset_password.rb
@@ -175,8 +175,10 @@ module Rodauth
       token_link(reset_password_route, reset_password_key_param, reset_password_key_value)
     end
 
-    def get_password_reset_key(id)
-      password_reset_ds(id).get(reset_password_key_column)
+    def get_password_reset_key(id, opts={})
+      ds = password_reset_ds(id)
+      ds = ds.where(Sequel::CURRENT_TIMESTAMP < reset_password_deadline_column) if opts[:ignore_expired]
+      ds.get(reset_password_key_column)
     end
 
     def login_form_footer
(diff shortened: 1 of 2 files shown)


jeremyevans:
Thanks for this patch. I agree with your analysis.  Your fix would change the get_password_reset_key API to support an optional argument.  That's one of the configuration methods, and I would like to avoid API changes there, so I think a different approach would be to delete any expired key before retrieving the key.    It is an extra query on every key retrieval, but this shouldn't be a bottleneck, and I think that is the safest approach and it doesn't require API changes.

jeremyevans:
I'll be releasing 1.12.0 today with this change, since I consider this a security fix.

Also, looks like I didn't address your second issue.  I don't consider that a bug, as raising errors for invalid input is a reasonable behavior.  That's not to say the behavior couldn't be improved in the future.  However, we can't just assume the account id column is integer, strings, UUIDs, and potentially other types could be used.

chanks:
Cool, thanks!

I understand your point about the error. I feel like that any input resulting in a 500 is a bug that needs fixing, but yeah, it's not a simple fix.

jeremyevans:
1.12.0 was just released with a fix for this issue.  Thanks for your help!

PR #45, opened 2017-10-21
Enable generic errors for login problems

fran-worley:
There are cases where you don’t want to tell the user what caused their
login to fail. This PR enables you to define a generic error message
and status to use in the event for login failure.

Its use is set to false by default for backwards compatibility.

### DIFF
--- a/doc/login.rdoc
+++ b/doc/login.rdoc
@@ -15,6 +15,9 @@ login_form_footer :: A message to display after the login form.
 login_notice_flash :: The flash notice to show after successful login.
 login_redirect :: Where to redirect after a sucessful login.
 login_route :: The route to the login action. Defaults to +login+.
+use_generic_login_errors? :: Use the same message and status for all login errors. Defaults to +false+.
+generic_login_error_status :: The status to use for login errors (only applicable if use_generic_login_errors? is set to true). 401 by default.
+generic_login_error_message:: The message to use for login errors (only applicable if use_generic_login_errors? is set to true). Defaults to +'there was a problem with the login or password supplied'+.
 
 == Auth Methods
 
--- a/lib/rodauth/features/login.rb
+++ b/lib/rodauth/features/login.rb
@@ -12,6 +12,9 @@ module Rodauth
 
     auth_value_method :login_error_status, 401
     auth_value_method :login_form_footer, ''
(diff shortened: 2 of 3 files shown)


jeremyevans:
Thanks for sending in a patch.

I believe you can accomplish what you want via the following rodauth configuration code:

```ruby
message = 'there was a problem with the login or password supplied'
no_matching_login_message message
unverified_account_message message
invalid_password_message message

status = 401
unopen_account_error_status status
login_error_status status
no_matching_login_error_status status

after_login_failure do
  @password_param = login_param
  super()
end
password_param do
  @password_param || super()
end
```

I don't think trying to be generic here is actually helpful from a security perspective.  It is usually simple for an attacker to determine whether a login is valid or not.  Even if you are not leaking the information via some other output, if you are using the default bcrypt for password hashing, a timing attack should be fairly trivial.

Because being generic is not something I want to encourage, and because it is already possible to do so via existing configuration code, I'm currently against merging this.  Are you OK with using the above configuration code?

fran-worley:
I'm actually using the something like the above already. The issue I have is that the error message variables are reused for other features (change password for example) where you'd want to have a specific message.

I take your point on the security aspects of 'generic' messages, but that decision sadly is out of my hands on this project and it seems like a significant number of web apps use generic messages, hence me putting this patch together.

jeremyevans:
You can scope these changes to just the login route:

```ruby
before_login_route do
  @generic_error_message = 'there was a problem with the login or password supplied'
  @generic_error_status = 401
  super()
end
[:no_matching_login_message, :unverified_account_message, :invalid_password_message].each do |meth|
  send(meth) do
    @generic_error_message || super()
  end
end
[:unopen_account_error_status, :login_error_status, :no_matching_login_error_status].each do |meth|
  send(meth) do
    @generic_error_status || super()
  end
end

after_login_failure do
  @password_param = login_param
  super()
end
password_param do
  @password_param || super()
end
```

I really don't want to encourage the use of generic messages, they are actively harmful to usability and provide no real security benefit.

Do you think the above configuration would work for you?

jeremyevans:
Closing due to lack of response.  If you find that there is something you need that you cannot do with the existing configuration code, please respond.

PR #46, opened 2017-10-26
Allow cache setting to be overridden

adam12:
Add's a `cache_templates` method to base, with a default value of true, allowing the caching of templates to be configurable.

Stops the clobbering of the `cache` value if provided via `template_opts`.

Uses `template_opts` as the base for `button_opts`.

### DIFF
--- a/doc/base.rdoc
+++ b/doc/base.rdoc
@@ -37,6 +37,7 @@ account_select :: An array of columns to select from +accounts_table+. By
                   default, selects all columns in the table.
 account_status_column :: The status id column in the account model.
 account_unverified_status_value :: The representating unverified accounts.
+cache_templates :: Whether to cache templates. True by default.
 default_redirect :: Where to redirect after most successful actions.
 invalid_field_error_status :: The response status to use for invalid field
                               value errors, 422 by default.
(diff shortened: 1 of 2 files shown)


jeremyevans:
This looks good.  I'll probably want to add a spec before merging.  I'm not sure if we want to check the existing template_opts :cache setting, we may just want to use the cache_templates method value unconditionally. 

Just so I understand, is the issue you are having is you want to update rodauth's template files during application development and see changes without restarting?  I would assume that modifying rodauth configuration during development would still pick up changes even with caching enabled, as long as you weren't modifying the template files themselves.  Or are you doing something else?

adam12:
Your understanding of my issue is correct.

I probably should of disclosed that I believe I'm relying on a small mis-feature of Rodauth, in that if I place a same-named view file in my own `views` directory, it's used instead of the Rodauth-packaged view.

So I'm basically bypassing the `*_view` rodauth methods. 😱

jeremyevans:
That's not actually a mis-feature, that's by design.  You aren't bypassing `*_view`, that's handled inside `*_view` (actually inside `_view_opts`).  It makes sense if you are doing that that you would want to turn caching off.  I think in that case, it can be expected you would set cache_templates to false in your rodauth configuration (or `cache_templates(ENV['RACK_ENV'] != 'development'`).

adam12:
> I think in that case, it can be expected you would set cache_templates to false in your rodauth configuration (or cache_templates(ENV['RACK_ENV'] != 'development').

My thoughts exactly. Thanks.

PR #49, opened 2018-06-02
Update "model instance" references

### LINKED ISSUE #48: References to "model instance"
I've noticed that Rodauth documentation still references "model instances" for methods like `#account_from_login` and `#account_from_session`. I remember that earlier Rodauth versions had `Sequel::Model` subclasses and created instances from it, but that it was changed at some point to work with hashes directly.

In that case I was wondering we should use difference terminology instead of "model instances". I wanted to make a PR, but I didn't know what would be the appropriate name. I was thinking "records", but I feel like it isn't suitable for `#new_account` because there the instance isn't a database record yet.

janko-m:
Rodauth doesn't use model instances anymore, it uses plain hashes instead. So, we update some leftover references of "model instances".

Closes #48

### DIFF
--- a/README.rdoc
+++ b/README.rdoc
@@ -618,7 +618,7 @@ response :: RodaResponse instance
 scope :: Roda instance
 session :: session hash
 flash :: flash message hash
-account :: account model instance (if set by an earlier Rodauth method)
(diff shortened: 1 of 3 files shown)


jeremyevans:

PR #51, opened 2018-08-04
Simplify expanding path

janko-m:
I noticed that expanding the default path for the top common passwords could be written in a shorter way.


jeremyevans:
Thanks for the patch.  This looks fine, I'll test and merge tomorrow.

PR #52, opened 2018-08-05
Fix typo in #authenticated? docs

janko-m:
Since the second part of the sentence says that both factors have to been authenticated, I think the first part meant to say "If 2 factor authentication *has* been enabled..."


jeremyevans:
Thanks for the patch!

PR #54, opened 2019-02-15
replaced hard-coded string 'already an account with this login' 

1gor:
Replaced hard-coded string 'already an account with this login' in several features with a single configurable `auth_value_method` in 'base' feature.

Some end-of-line spaces were automatically removed by emacs.

Added rules  to `.gitignore` to ignore files often created by a development environment.

### DIFF
--- a/doc/login_password_requirements_base.rdoc
+++ b/doc/login_password_requirements_base.rdoc
@@ -5,6 +5,9 @@ use a Rodauth feature that requires setting logins or passwords.
 
 == Auth Value Methods
 
+already_an_account_with_this_login_message :: The error message to display when
+                                              there already exists an account
+                                              with the same login
 login_confirm_label :: The label to use for login confirmations.
 login_confirm_param :: The parameter name to use for login confirmations.
 login_does_not_meet_requirements_message :: The error message to display when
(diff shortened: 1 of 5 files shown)


jeremyevans:
Thanks for working on this.  I agree that the `already_an_account_with_this_login_message` addition is a good idea. Please push a version with the following changes:

1. Do not modify `.gitignore` (you can use `.git/info/exclude` for local ignores)
2. Add to login_password_requirements_base feature instead of base feature
3. Avoid modifying whitespace on lines where content doesn't change
4. Make sure the `auth_value_method` call is in alphabetical order
5. Update `doc/login_password_requirements_base.rdoc`

1gor:
Done.

My editor still automatically removes trailing spaces from the code (such as line [line 106](https://github.com/jeremyevans/rodauth/pull/54/commits/8cbd8041346dcde2e37734100485f919f2a3d343#diff-5ae014e7752944ea1834ca9c18b0c944R106) in `lib/rodauth/features/create_account.rb`).

It takes time to revert those space removals while committing. To change my editor setting every time I work on this particular repo is also quite inconvenient. Do you think leaving trailing spaces as they are in comments or empty lines is essential? Maybe you could accept them in future commits as those changes arguably clean up the code somewhat?

jeremyevans:
Thanks, the changes look good.  I will test and merge these changes on Tuesday.

It's not that leaving whitespace at the ends of lines is important, but not doing so makes git blame harder to use.

I don't want to force all users to use an editor that strips whitespace at the ends of lines, or reject pull requests that add whitespace at the ends of lines.  Whitespace at the ends of lines is unimportant, the important parts are the code changes.  I don't consider removing whitespace at the ends of lines as "cleaning up the code".

PR #55, opened 2019-02-19
Implemented 'login return to' feature

1gor:
This feature implements a common use case when a user navigates to a protected url, is shown a login page, logs in successfully and expects then to be shown the initially requested page.

Without this feature a user wishing to reach, for example, the protected "/profile" page will be presented with a login form and then sent to the home page rather than to his target url.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,7 @@
+=== master
+
+* Add already_an_account_with_this_login_message configuration method (1gor) (#54)
+
 === 1.19.1 (2018-11-16)
 
 * Support rotp 4 in the otp feature (jeremyevans)
--- a/doc/login_return_to.rdoc
+++ b/doc/login_return_to.rdoc
@@ -1,0 +1,9 @@
+= Documentation for Login Return To Feature
+
+The login return to feature allows to remember the requested protected url and
+to redirect back to it after successful login. It depends on login feature.
+
+== Auth Value Methods
+
+login_return_to_session_key :: session key name that holds the
+                               requested url, defaults to 'return_to'.
--- a/lib/rodauth/features/login_return_to.rb
+++ b/lib/rodauth/features/login_return_to.rb
@@ -1,0 +1,26 @@
+# frozen-string-literal: true
+
+module Rodauth
+  Feature.define(:login_return_to, :LoginReturnTo) do
+    depends :base, :login
+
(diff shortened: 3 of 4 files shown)


jeremyevans:
Thanks for the pull request.  I've implemented something similar to this in a couple of different applications.  It seems common enough to be worth supporting in Rodauth. 

One reason I haven't added it to Rodauth myself is in all of my applications where I have supported something similar to this, I have added checks to protect against the use of bad paths.  I think the checks I've added are sufficient for my applications, but they may be too restrictive in general, and I'm not 100% sure they cover all bad cases.

Blinding trusting any path provided by the user seems to be a bad idea to me.  Without some sort of path validation, I think there may be a risk of some form of header injection.  Have you considered the security implications of this change, and if so can you provide some evidence supporting the security of this approach?

jeremyevans:
Another thing I should add is that storing this information in the session allows a user to easily break the session using a long path/query string (assuming you are using cookie sessions).  In applications where I have supported this, I store the redirect path in a hidden input field on the login form.

1gor:
One way to mitigate security issues is to add a regex to this plugin's configuration, something like `auth_value_method :valid_redirect_paths, /^\/[(foo|bar),0-9,\/]{0,10}$/`. That would match '"/foo/1/bar" (or other paths that the developer would allow) and limit the length of the string to protect the session.  We could offer some sensible default (and I am not a regex expert by any means, so will need some help here).

Admittedly, this shifts the security burden to the developer. But I think this approach is in the spirit of rodauth itself that offers both tools and flexibility. Otherwise I have a hard time thinking of how to whitelist application paths given that roda does not offer any route introspection...

jeremyevans:
A regexp approach is what I used as well.  I think my regexp checks the path doesn't have any whitespace.  So I think that we should force a regexp.  I'm not sure what that regexp should be by default.  Not having whitespace is probably OK, but I'm open to a more conservative approach.  If a user doesn't want to validate, they could use the empty regexp.

After that change, I think this implementation would just need to switch so that the redirect path was not stored in the session but in a hidden input on the login form.  The tests should probably include a test where login fails the first time before succeeding, and also when `use_multi_phase_login?` is set to true.

1gor:
I will try to implement your suggestions when I have time and am closing this pull request for now.

PR #56, opened 2019-02-23
Update copyright notice to 2019

Atul9:


jeremyevans:
Thanks for the patch. This makes sense, I've made changes in 2019.

PR #57, opened 2019-03-31
Disable autocomplete on OTP Auth Code input

adam12:
This code is likely input from a separate device and any previous values are unnecessary noise. 

Adds an `attr` option to the `opts` hash passed to `input_field_string`, allowing us to pass arbitrary attributes.


jeremyevans:
Thanks, this looks good and I agree it would be useful.

PR #59, opened 2019-05-03
Fix disallow password reuse feature when not using verify_account

cptaffe:
When account_password_hash_column is unset and the verify account feature is not loaded, verify_account_set_password? is called raising a NoMethodError.


jeremyevans:
Thank you very much for the patch.  I'll add a spec for this and merge your change later today.

Issue #60, opened 2019-06-05
Unneeded access_token for incorrect password login


xvonabur:
Current implementation of `jwt_refresh` feature returns `access_token` and `Authorization` header for incorrect password case. Example of HTTP body:

```JSON
{"access_token":"eyJhbGciOiJIUzI1NiJ9.eyJleHAiOjE1NTk3NTI0NzUsImlhdCI6MTU1OTc1MDY3NSwibmJmIjoxNTU5NzUwNjcwfQ.FXz9tQb3jHuxnuXQ5QUt5q0ZBTP53LGTHCz4FfaMu98",
"field-error":["password","invalid password"],
"error":"There was an error logging in"}
```

I think we need to delete them in case of any error. Otherwise, anyone can get `access_token` based on only existent account login.

jeremyevans:
Have you tried using the provided access token?  It don't think it actually gives you access to anything. One thing to understand about rodauth's jwt support is that the jwt contains the same data as the session.  That's how rodauth is able to support jwt for all features without explicit jwt support in any of them.

xvonabur:
Yeap, just tested returned `access_token` and it doesn't provide access to authorized section of an app. Ok, so is there a way to omit it in case of error? I'm trying to build an API with `json_only` mode, so it can be confusing for other developers to see an `access_token` in error case.

jeremyevans:
Maybe:

```ruby
  set_jwt_token do |token|
    super(token)
    if json_response[json_response_error_key]
      json_response.delete(jwt_access_token_key)
      response.headers.delete('Authorization')
    end
  end
```

xvonabur:
@jeremyevans Thanks! I've added it to configuration block and it solves my problem. Can we add this as default behaviour for `jwt` feature at least?

jeremyevans:
Deleting the jwt_access_token_key in set_jwt_token inside the jwt_refresh feature is probably acceptable.  That feature is new and there are no backwards compatibility issues.  Plus, the code is already setup not to add to the response if there is an error, so deleting from the response if there is an error is a reasonable change.  Can you send a pull request for that?

I don't think deleting the header makes sense, since that is done by the jwt feature.  As you saw, it doesn't cause any security issues.  It is actually recommended to use the new headers/jwt even in case of an error.  Basically, every request that returns a jwt in the header, that jwt should be used for future requests until a request returns a jwt in the header.

xvonabur:
> Can you send a pull request for that?

[...]

PR #61, opened 2019-06-06
Omit jwt access_token from response on error

xvonabur:
Implemented feature from #60


jeremyevans:
Thanks for the patch. This looks good, I'll try to test and merge later today.

PR #64, opened 2020-02-15
Add *_path and *_url counterparts to *_route methods

janko:
Having a relative or absolute URL to the action is useful when redirecting or displaying links. They are also used in a few places inside the Rodauth codebase, which would be useful to DRY up.

So we add `*_path` and `*_url` counterparts to each `*_route` method, which return relative and absolute URL to the route respectively.

```rb
login_route #=> "login"
login_path  #=> "<prefix>/login"
login_url   #=> "https://example.com<prefix>/login"
```

### DIFF
--- a/doc/change_login.rdoc
+++ b/doc/change_login.rdoc
@@ -27,4 +27,6 @@ before_change_login_route :: Run arbitrary code before handling a change login r
 change_login(login) :: Change the users login to the given login, or
                        return nil/false if the login cannot be changed to
                        the given login.
+change_login_path :: The relative URL to the change login action.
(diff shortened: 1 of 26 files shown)


jeremyevans:

jeremyevans:
Looks like specs are failing because login_path is being used in a place where it may not be defined.  I'll fix that.  I also think we shouldn't make `*_path` and `*_url` auth methods, since there is no reason to override those separately.  I'll fix that as well.

janko:
Thanks a lot for the fixes! That makes much more sense, and that way the docs are more DRY 👍

PR #65, opened 2020-02-15
Add #valid_login_entered? method for multi phase login

janko:
When using the email_auth feature, we want to display the option to authentication via email when the user has entered a valid login during multi phase login.

By default, the email_auth feature overrides `#login_form_footer` to add the form for requesting email authentication. However, in my case I'm rewriting Rodauth views in Rails, and don't want to use Rodauth's view helpers such as `#login_form_footer` (I want all HTML to be contained in views).

So, I added the `#valid_login_entered?` method that reads the instance variable, which can then be called in views like this to display the email auth request form on the login page:

```rb
# ... login form ...

if rodauth.valid_login_entered?
  rodauth.email_auth_request_form
end
```

We also get rid of the `@email_auth_request_form` variable, which doesn't seem to be necessary (but I could be wrong). This changes the behaviour to display the email auth request form even when the user has entered both login and password, but the password was invalid, which I think is desirable (the user might have forgotten their password, so it's useful for them to see the alternative of email authentication).

### DIFF
--- a/lib/rodauth/features/email_auth.rb
+++ b/lib/rodauth/features/email_auth.rb
@@ -150,7 +150,7 @@ module Rodauth
 
     def login_form_footer
       footer = super
-      footer += @email_auth_request_form if @email_auth_request_form
+      footer += email_auth_request_form if valid_login_entered?
       footer
     end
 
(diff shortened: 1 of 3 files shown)


jeremyevans:
Thanks for the patch. I'm OK with the change, but can you update the email_auth spec to test the described change in behavior?

janko:
Good suggestion, I've added the spec, and have verified that it fails before this change.

jeremyevans:
Thanks, spec looks good.  I'm currently traveling to a conference, but I'll test and merge this when I return, probably in a little over a week.

janko:
Sounds good, this will give me the time to test my Rails integration and make sure everything is working correctly. Enjoy Paris.rb!

PR #66, opened 2020-02-15
Reuse #login_hidden_field method

janko:
I noticed the "email-auth-request-form" and "login-display" templates use the `#login_hidden_field` method, while other templates that use a login hidden field enter the HTML manually. This changes other templates to also use the `#login_hidden_field` method for consistency.


jeremyevans:
Thanks for the patch.  This looks fine as all templates are for features that depend on the login feature (which defines login_hidden_field).

jeremyevans:
Merged as d05c0a9cf8cee2e4c0f8a730dc058aadb689007c

PR #69, opened 2020-04-04
Allow redirecting to originally requested URL after requiring authentication

janko:
When the user visits a certain page, but this page requires authentication, it's nice UX if the user is redirected back to the page they originally requested after successful authentication. This pull request adds this feature as discussed on the google group.

We implement this by storing the current location into a session on `login_required` and `require_two_factor_authentication`, and then redirecting back to that location on successful authentication, removing the session value in the process. I chose the "return to" naming as that's what Devise and Clearance use. ~We store the full URL to handle potential subdomains.~

This works with email authentication and webauthn login as well. We use separate settings for login and two factor authentication, as two factor authentication doesn't require the login feature to be loaded, and the user might want to configure these separately anyway (e.g. redirect to two factor auth directly after login, but have two factor authentication remember requested URL).

We don't override `login_redirect` and `two_factor_auth_redirect` so that they can still be used for configuring default redirects for when the user manually visited the authentication page, while still keeping the new functionality.

For login we need to load the session key into an instance variable because `#login_session` clears the session. For that reason I've removed the `#clear_session` call before `POST /login`, which apparently didn't break any tests.

Even though Rodauth 2.0 is on the horizont, I've chosen to disable this feature by default after all, because it could lead to unexpected redirects in scenarios like the following:

1. user browses a website
2. they encounter a page that requires authentication
3. they go back an continue browsing the website some more
4. at some point they decide to manually login in / create account
5. they're redirected to the previously stored URL that was never cleared

This decision also allows us to avoid breaking existing tests that aren't counting with this functionality.

I considered solving this using flash messages that would be discarded when user continues browsing the app (i.e. visits the non-rodauth routes), and thus would reduce the chance for an unexpected redirect. However, it would introduce backwards compatibility for users iterating over all flash keys assuming they're all either message that need displaying. It would also introduce complications with the JWT feature which doesn't load flash by default.

### DIFF
--- a/doc/login.rdoc
+++ b/doc/login.rdoc
@@ -13,6 +13,8 @@ login_form_footer :: A message to display after the login form.
 login_need_password_notice_flash :: The flash notice to show during multi phase login after the login has been entered, when requesting the password.
 login_notice_flash :: The flash notice to show after successful login.
 login_redirect :: Where to redirect after a sucessful login.
+login_return_to_requested_location? :: Whether to redirect to the originally requested location after successful login when require_login was used, false by default.
+login_redirect_session_key :: The key in the session hash storing the location to redirect to after successful login.
 login_route :: The route to the login action. Defaults to +login+.
 use_multi_phase_login? :: Whether to ask for login first, and only ask for password after asking for the login, false by default.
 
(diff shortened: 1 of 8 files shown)


jeremyevans:
Thanks for the patch! This looks good, I'll try to test and merge on Monday.

janko:
As suggested in https://github.com/jeremyevans/rodauth/pull/75#issuecomment-609115984, I updated the PR to store only the path instead of the absolute URL. I've also updated tests to verify that we're also retaining any query parameters.

PR #70, opened 2020-04-04
Fix duplicate spec name

janko:
I noticed this spec name is the same as the first spec in this file, so I updated the description.


jeremyevans:
Thanks for the patch!

PR #71, opened 2020-04-04
Use webauth_login_failure_redirect on webauthn login failure

janko:
This was most likely just not updated after introducing the redirect option.


jeremyevans:
Thanks for the patch!

PR #72, opened 2020-04-04
Inline create_account dependencies

janko:
This puts `create_account` feature dependencies into the same `depends` call.


jeremyevans:
Thanks for the patch!

PR #73, opened 2020-04-04
Specify the JwtRefresh constant in the Feature.define call

janko:
This makes jwt_refresh feature definition consistent with all other features.


jeremyevans:
Thanks for the patch!

PR #74, opened 2020-04-04
Remove unused view render in login feature

janko:
It noticed a view render that doesn't seem to be used anywhere, so I thought it's probably a leftover from a previous refactor.


jeremyevans:
You are correct, good catch!

PR #75, opened 2020-04-04
Add #require_password_confirmation for easy "sudo" mode

janko:
GitHub has a "sudo" mode for sensitive operations, where they redirect to a page where the user can confirm their password before continuing. They also have a grace period, meaning password confirmation is not required if password was entered recently.

The confirm_password feature already provides tools for building this functionality, but it requires a few method calls. So we add a `#require_password_confirmation` method that will redirect to the password confirmation page, then redirect back to the original page. It also respects the password_grace_period feature.

### DIFF
--- a/README.rdoc
+++ b/README.rdoc
@@ -966,6 +966,13 @@ require_current_password :: (password_expiration feature) Require a current
                             password, redirecting the request to the change
                             password page if the password for the account has
                             expired.
+require_password_confirmation :: (confirm_password feature) Redirect to the
+                                 password confirmation page, saving the current
+                                 location to redirect back to after password
+                                 has been successfully confirmed. This should
+                                 be wrapped in a conditional that evaluates
+                                 to false after password confirmation to avoid
+                                 endless redirects.
 load_memory :: (remember feature) If the session has not been authenticated, look
                for the remember cookie.  If present and valid, automatically
                log the session in, but mark that it was logged in via a remember
--- a/doc/confirm_password.rdoc
+++ b/doc/confirm_password.rdoc
@@ -18,6 +18,9 @@ confirm_password_notice_flash :: The flash notice to show after password confirm
 confirm_password_redirect :: Where to redirect after successful password confirmation. By default, uses <tt>session[confirm_password_redirect_session_key]</tt> if set, allowing an easy way to redirect back to the page requesting password confirmation.
 confirm_password_redirect_session_key :: The session key used to check for the confirm_password_redirect.
 confirm_password_route :: The route to the confirm password form. Defaults to +confirm-password+.
+need_password_confirmation_error_status :: The response status to use if going to a page requiring password confirmation, 401 by default.
+need_password_confirmation_error_flash :: The flash error to show if going to a page requiring password confirmation.
+password_confirmation_required_redirect :: Where to redirect when going to a page requiring password confirmation.
 
 == Auth Methods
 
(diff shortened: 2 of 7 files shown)


jeremyevans:
This is interesting, but I'm not sure how it works without the password_grace_period feature.  Let's say you had this enabled for a certain route/branch.  If you don't have a grace period, wouldn't it continually redirect back to the confirmation page after submitting the confirmation page?

Both for this and #69, I think it is best to only store the request path in the session and not the full URL.  This can prevent cross domain redirect attacks if the app can be called with any Host header.

I'm definitely open to this idea as an alternative to the password on ever page design, but it should probably be part of the password_grace_period feature or at least depend on it.  Maybe password_grace_period could depend on the confirm password feature so that it is always available?

Finally, methods that the user can be expected to call in their app's routing blocks should always have tests.  Rodauth doesn't have tests for usage of all configuration methods as the design doesn't make that necessary, but for things the user is expected to call directly, we should have tests to help ensure correct behavior and prevent regressions.

janko:
Oh, you're right, it would just result in a endless redirect without password_grace_period loaded. Ok, I will move it to the password_grace_period feature instead, and make password_grace_period depend on confirm_password.

I will also rename it to `require_password_recently_entered`, as that communicates that the password confirmation will not always be required, and that it's connected to the `password_recently_entered?` method (and I can now call that method directly on the conditional).

janko:
Ok, I made the changes, and have renamed other configuration options accordingly (so that they don't indicate that they're coming from the `confirm_password` feature). I've also added tests for JWT.

janko:
Renamed it back to `#require_password_confirmation`, as that communicates where the request will redirect to, and I just like how it reads.

janko:
After more consideration, I've decided the add `#require_password_confirmation` back to the confirm_password feature, but without any conditionals. The user can then compose the behaviour they want using conditionals from other features. If they want to require password confirmation if the password hasn't been entered recently, they can do it with:

```rb
unless rodauth.password_recently_entered?
  rodauth.require_password_confirmation
end
```

If they want to require password confirmation for users autologged in from a remember token, they can do it with:

```rb
if rodauth.logged_in_from_remember_key?
  rodauth.require_password_confirmation
end
```

This makes it easier to require password confirmation for remember feature, and it doesn't require the password_grace_period feature to depend on confirm_password. I added a note in the README that this method should be wrapped in a conditional that evaluates to false after password confirmation to avoid endless redirect.

To enable usage with password_grace_period, I've made the `#password_recently_entered?` method public. Theoretically `#modifications_require_password?` method can also be used, which is already public, but `#password_recently_entered?` is more explicit and makes it clear that there is a grace period.

Let me know what you think.

jeremyevans:
This would make `require_password_confirmation` behavior different from other similarly named methods.  `require_login`, `require_authentication`, `require_two_factor_setup`, `require_current_password` all redirect to a page where after you successfully complete the action, the method will no longer redirect.  If you want to always redirect, I think it would be better to use a method name that explicitly specifies redirection.

I think it would be better to make the behavior consistent, so that you can just use `rodauth.require_password_confirmation` directly.  By default, it should redirect if `authenticated_by` does not include `password`.  The password_grace_period feature can change it so it also redirects if password was not entered recently.  That should handle both the cases you are mentioning (remember and password grace period), as well as other cases (email_auth and webauthn_login). I think I can take care of making this change after merging.

janko:
Agreed, I also didn't like that it didn't match the existing `require_*` methods. Your idea sounds perfect 💯

PR #76, opened 2020-04-04
Redisplay the reset password form in validation errors

janko:
This redisplays reset password request form in case of validation errors, allowing the user to correct their login if they mistyped it.

In order to do this, I needed to avoid displaying a hidden login field in the form when a login has been entered. I initially thought about just replacing the conditional in the form to use `rodauth.valid_login_entered?` instead, but that is only true when `use_multi_phase_login?` is true as well, so when `use_multi_phase_login?` is false we would be displaying a reset password request form with an email field as well, which is probably not desired.

Since we already have the "Forgot Password?" link always displayed in the login footer, I thought we could get away without display another reset password request button in the login header as well. Yes, the "Forgot Password?" link requires the user the re-type their login, but I think that's a fair trade-off for reduced complexity.

(I initially intended to do this for the email_auth feature as well, but I realized that the email auth request form is currently only displayed as part of a multi-phase login, and doesn't have a `GET` route, so we're good there.)

### DIFF
--- a/doc/reset_password.rdoc
+++ b/doc/reset_password.rdoc
@@ -49,7 +49,6 @@ create_reset_password_key :: Add the reset password key data to the database.
 create_reset_password_email :: A Mail::Message for the reset password email.
 get_reset_password_email_last_sent :: Get the last time a reset password email is sent, or nil if there is no last sent time.
 get_reset_password_key(id) :: Get the password reset key for the given account id from the database.
-login_failed_reset_password_request_form :: The HTML to use for a form to request a password reset, shown on the login page after the user tries to login with an invalid password.
 remove_reset_password_key :: Remove the reset password key for the current account, run after successful password reset.
 reset_password_email_body :: The body to use for the reset password email.
 reset_password_email_link :: The link to the reset password form in the reset password email.
--- a/lib/rodauth/features/login.rb
+++ b/lib/rodauth/features/login.rb
@@ -68,8 +68,6 @@ module Rodauth
       end
     end
 
-    attr_reader :login_form_header
-
     def after_login_entered_during_multi_phase_login
       set_notice_now_flash need_password_notice_flash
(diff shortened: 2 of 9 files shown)


jeremyevans:
This looks good.  I'll merge and test on Monday.  The only issue I can see with it is that `login_failed_reset_password_request_form` should be removed from the reset password documentation.  However, I can fix that after merging.

janko:
Thanks, I saw that I left out the `login_failed_reset_password_request_form` method definition as well, so I removed both now.

jeremyevans:
After some more thought, I want to keep the direct reset password request button for failed logins.  Forcing people to type in their email address a second time provides a poor user experience, and I think the benefit of the nicer experience outweighs the complexity cost.  So I'll merge the second commit, but not the first.

Only merging the second commit results in an error because it doesn't allow the user to correct a bad login.  However, that can be more easily be fixed by not using a hidden field if there is a field error.  I've fixed that locally.

janko:
Sounds reasonable, and good thinking about checking the field error.

Just a suggestion, but what do you think about still keeping everything in the footer, and having the reset password request form replace the "Forgot Password?" link in this case? So, something like this:

```rb
   def after_login_failure
     unless only_json?
       @reset_password_request_form = login_failed_reset_password_request_form
     end
     super
   end

    def _login_form_footer_links
      links = super
      if defined?(@reset_password_request_form)
        links << [20, @reset_password_request_form]
      else
        links << [20, reset_password_request_link]
      end
      links
    end
```

Though that would not really be compatible with https://github.com/jeremyevans/rodauth/pull/77.

jeremyevans:
I think it someone enters the correct username but incorrect password, requesting a password reset is probably their most likely action (unless they think the typoed the password), so having it above the standard login form makes sense.

Also, I already merged #77, so no going back now. :)

PR #77, opened 2020-04-04
Make login form footer links more consistent with two factor links

janko:
I noticed that login form footer links are implemented slightly differently than the two factor links, in a way that the whole link is stored instead of just link text. If this was done intentionally for backwards compatibility with existing `*_link` methods, please disregard this pull request.

In this pull request I extracted only link texts into value methods, which makes it easier for someone to write a custom template (as they can still read the link texts array but customize the link markup). I've also moved the login form footer into a template to simplify markup generation.


jeremyevans:
As you thought, the current design was done specifically for backwards compatibility with the existing configuration methods.  However, considering we are dropping some backwards compatibility in Rodauth 2, I will certainly consider this approach.  Let me take a day or two to think about this.

jeremyevans:
After some thought, I'm fine breaking backwards compatibility here to allow for greater consistency.  I'll test and merge this today.

PR #78, opened 2020-04-05
Use #convert_token_key method for converting remember token key

janko:
This shortens a bit converting the remember key by using the `#convert_token_key` method.


jeremyevans:
Thanks, this looks good.

PR #79, opened 2020-04-05
Remove confirm_password dependency for remember feature

janko:
The confirm_password feature was originally part of the remember feature, but now that it's extracted, it appears to me these two features are in fact decoupled from one another. If I'm mistaken, please disregard this pull request.

While it provides extra security to require password confirmation for sensitive parts of the app when logged in via remember token, for those that aren't doing this it's beneficial if they don't need to load the confirm_password feature code.

### DIFF
--- a/doc/remember.rdoc
+++ b/doc/remember.rdoc
@@ -1,10 +1,12 @@
 = Documentation for Remember Feature
 
-The remember feature allows for token-based autologin for users. It records
-which sessions were autologged in via a token, and allows you to request
-password confirmation later for such sessions if they are accessing a
-section requiring more security.  The remember feature depends on the
-logout feature.
+The remember feature allows for token-based autologin for users. Calling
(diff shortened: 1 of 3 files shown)


jeremyevans:
I don't think there is a hard dependency between the remember and confirm_password.    The documentation for remember implies features implemented by confirm_password, and the dependency exists for backwards compatibility for Rodauth pre-1.0.  I'm fine removing the dependency in Rodauth 2, but can you update the remember documentation to reflect that that you need to load confirm_password separately if you want to confirm passwords?

janko:
I've updated the docs to reference the confirm password feature. I didn't read so well when I tried explaining this in the first paragraph, so I decided to move it in a separate paragraph, and use the first one to explain the base usage of the remember feature.

The code example references a `rodauth.require_password_confirmation` method that I will add to the confirm_password feature shortly in https://github.com/jeremyevans/rodauth/pull/75.

PR #80, opened 2020-04-06
Remove remembered session key

janko:
Now that we're storing the authentication method in the `authenticated_by` session key, which we are using to check whether the current session has been autologged in via a remember token, I believe we don't need the `remembered` session key anymore.


jeremyevans:
Thanks, this looks good.  I'll merge and test this today.

PR #81, opened 2020-04-07
Fix checking #login_uses_email? for autocomplete

janko:
Fixes the condition for the login autocomplete value to call the `#login_uses_email?` method.


jeremyevans:
Good catch, thanks!

PR #82, opened 2020-04-08
Specify webauthn auth form action directly

janko:
In rodauth-rails, I'm rewriting Rodauth templates into Rails ERB views, and I found everything directly translatable, with the exception of the `webauthn_auth_form_attr` usage in the `webauthn-auth` template. This is because in Rails it's a given to use `form_*` helper method, which cannot really utilize a raw HTML attribute list without parsing.

I've noticed that the `webauthn_auth_form_attr` setting seems to only be used currently by the webauthn_login feature to set the form action. And since webauthn_login also overrides `webauthn_auth_form_path` with the same value, I thought we could just use that directly. I don't know if there were plans to utilize `webauthn_auth_form_attr` further than that.

This changes the form `action` value of a regular webauthn auth form without webauthn_login feature loaded to be explicitly set. I don't know if that changes behaviour.


jeremyevans:
This should be fine, there isn't really a problem with explicitly specifying the action in a form even when it is the same as the current path. I'll merge and test shortly.

jeremyevans:
Merged as 6cf7313c7877c72234d940930396068f78156217

PR #83, opened 2020-04-14
Make templates compatible with Bootstrap 4

janko:
This pull request updates templates to be compatible with Bootstrap 4, while retaining Bootstrap 3 compatibility as much as possible:

* add [server-side validation errors](https://getbootstrap.com/docs/4.4/components/forms/#server-side) using `is-invalid` and `invalid-feedback` classes
* use [`form-control-plaintext`](https://getbootstrap.com/docs/4.4/components/forms/#readonly-plain-text) class when displaying entered login during multi-phase login (and use `form-control-static` for Bootstrap 3)
* update radio buttons and checkboxes markup with inputs and labels being side-by-side instead of input being the child of the label (the latter is not mentioned anymore in Bootstrap 4 docs)
* wrap `col-*` divs with `row` divs to avoid an indent margin

As we've discussed on the google group, to avoid current alignment issues with horizontal forms, we switch to regular forms which makes it easier to have consistent design and reduces markup.

I've tried to maintain Bootstrap 3 compatibility, and everything looks good to me except that radio buttons and checkboxes have negative margins, which are added by the `radio` class. This class is needed to provide appropriate spacing, but unfortunately I wasn't able to cancel the negative margin because Boostrap 3 doesn't yet have margin utility classes.

I've also done some other related changes:

* extracted the global logout checkbox into a partial due to increased amount of markup
* used `btn-danger` for close account button, as I believe that's more appropriate
* removed `with_index` in the webauthn remove form since the index didn't seem to be used

### DIFF
--- a/lib/rodauth/features/active_sessions.rb
+++ b/lib/rodauth/features/active_sessions.rb
@@ -89,9 +89,7 @@ module Rodauth
     end
 
     def logout_additional_form_tags
-      input = input_field_string(global_logout_param, "rodauth-global-logout", :type=>'checkbox', :skip_error_message=>true, :value=>'t', :required=>false)
-      input = "<label class=\"rodauth-global-logout-label\">#{input} #{global_logout_label}</label>"
-      super.to_s + input
+      super.to_s + render('global-logout-field')
     end
 
     def update_session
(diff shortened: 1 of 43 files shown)


jeremyevans:
Thank you very much for working on this.  I hope to have time today to review and merge these changes.

PR #84, opened 2020-04-15
Allow passing query parameters to *_path and *_url methods

janko:
As discussed on the google group, this PR adds the ability for `*_path` and `*_url` methods to accept query parameters. This allows one to replace:

```rb
"#{rodauth.create_account_path}?#{Rack::Utils.build_nested_query(foo: "bar")}"
#=> "/create-account?foo=bar"
```

with

```rb
rodauth.create_account_path(foo: "bar")
#=> "/create-account?foo=bar"
```

I needed this in my application where I wanted to customize the registration form depending on the context, which I'm passing as query parameters. And Rails didn't have any helper methods for adding query parameters to a path, so I thought it would be useful to Rodauth to have that built in.

### DIFF
--- a/README.rdoc
+++ b/README.rdoc
@@ -1029,9 +1029,11 @@ authenticated_webauthn_id :: (webauthn feature) If the current session was
                              authenticated via webauthn, the webauthn id of the
                              credential used.
 *_path :: One of these is added for each of the routes added by Rodauth, giving the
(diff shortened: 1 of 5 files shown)


jeremyevans:
Thanks for the patch.  This looks good, I'll test and merge soon.

PR #85, opened 2020-04-16
Ensure webauthn-remove error message is displayed on Bootstrap 4

janko:
As described [here](https://groups.google.com/d/msg/rodauth/j3Qc2N1R1MU/xikp41F8AQAJ), the Bootstrap 4 change introduced a regression regarding displaying the validation error message.

Bootstrap 4 requires the input field to have class `is-invalid` and to be a sibling to the `invalid-feedback` element, otherwise the `invalid-feedback` element won't be displayed.

Since the webauthn remove field error will only be displayed if no options were selected, we display the error message after the last radio button. We also use the `#input_field_string` method to get the `is-invalid` class back, but modify it to allow for checkboxes as well.

The tests are failing with this change, it seems the radio buttons are not being displayed anymore on the validation error.

### DIFF
--- a/lib/rodauth/features/base.rb
+++ b/lib/rodauth/features/base.rb
@@ -175,6 +175,8 @@ module Rodauth
         value = opts.fetch(:value){scope.h param(param)}
       end
 
+      field_class = opts.fetch(:class, "form-control")
+
       if autocomplete_for_field?(param) && opts[:autocomplete]
         autocomplete = "autocomplete=\"#{opts[:autocomplete]}\""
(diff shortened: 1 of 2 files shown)


jeremyevans:
Thanks for the patch. I'll look at this today and see if I can get the visual issues fixed while still keeping the tests passing.

jeremyevans:
I'm just going to remove the error message check in this case.  It can only be hit in a race condition, and I don't think it's worth trying to handle that race condition.

PR #86, opened 2020-04-21
Improvements to the http_basic_auth feature

janko:
As discussed on the google group, this introduced couple of fixes and improvements for the http_basic_auth feature.

First, it renames the boolean `#require_http_basic_auth` method to `#require_http_basic_auth?` for consistency with other boolean auth value methods.

Next, I wanted to call `#login_session` when logging in, so that things like `#authenticated_by` are properly set which other features rely on. However, I couldn't do so when we were overriding `#session`, because `#login_session` calls `#session` internally, which created an endless loop.

So I decided to extract this behaviour into a `#http_basic_auth` method. Now, instead of HTTP basic authentication being allowed for any request that calls `#session`, the developer needs to call `#http_basic_auth` in their routing tree to get that behaviour (I find this nicer as it's more explicit):

```rb
route do |r|
  rodauth.http_basic_auth
  r.rodauth
  # ...
end
```

Lastly, I've extracted requiring HTTP basic authentication into a new `#require_http_basic_auth` method, which IMO fits nicely with other `#require_*` methods. Now the developer can call `#require_http_basic_auth` when they want to require HTTP basic authentication, and `#http_basic_auth` when they want to allow it but not require it.

I should probably write more thorough specs for `#http_basic_auth`; e.g. the way it's currently written it allows re-logging in without logging out first, which could be a useful feature that someone might rely on in the future. I should also write specs for `#require_http_basic_auth`, as it's a public-facing method.

Firstly I wanted to check with you whether you agree with this direction.

### DIFF
--- a/doc/http_basic_auth.rdoc
+++ b/doc/http_basic_auth.rdoc
@@ -6,4 +6,4 @@ described in RFC 1945.
 == Auth Value Methods
 
 http_basic_auth_realm :: The realm to return in the WWW-Authenticate header.
-require_http_basic_auth :: If true, when +rodauth.require_authentication+ is used, return a 401 status if basic auth has not been provided, instead of redirecting to the login page. False by default.
+require_http_basic_auth? :: If true, when +rodauth.require_authentication+ is used, return a 401 status if basic auth has not been provided, instead of redirecting to the login page. False by default.
--- a/doc/http_basic_auth.rdoc
+++ b/doc/http_basic_auth.rdoc
@@ -3,6 +3,10 @@
 The HTTP basic auth feature allows logins using HTTP basic authentication,
 described in RFC 1945.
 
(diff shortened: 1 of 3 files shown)


jeremyevans:
I agree with this direction.  Thank you for working on this.

janko:
@jeremyevans Happy to help 😃. Specs updated ✅ 

I've decided to allow re-HTTP-basic-authenticating without first logging out for both `#http_basic_auth` and `#require_http_basic_auth`, just so that people using `#require_http_basic_auth` can still use that feature. I'm not sure how common it would be, but I didn't see any reason to not support it (and it's one less `#logged_in?` check 😛).

jeremyevans:
Thanks, this looks good.  I'm going to merge this and then make some modifications.

janko:
Thank you, and as always, your edits make total sense to me 👌

PR #87, opened 2020-04-21
Don't call #check_already_logged_in in verify_login_change

janko:
When verifying login change, the user is expected to be logged in, because prior to that they've changed their login, which required them to be logged in. This action also does generally make sense when the user is logged in.

In my case, this allows me to configure

```rb
already_logged_in { redirect login_redirect }
```

and have that be called for login, create_account and reset_password, but still allow access to the verify_login_change page.

### DIFF
--- a/lib/rodauth/features/verify_login_change.rb
+++ b/lib/rodauth/features/verify_login_change.rb
@@ -50,7 +50,6 @@ module Rodauth
     )
 
     route do |r|
-      check_already_logged_in
       before_verify_login_change_route
 
       r.get do


jeremyevans:
One thing I worry about when allowing this is possible avenues of attack when going to the very login change link for one account while logged into a different account.  There may be no problems in doing so, but I'd need to check to make sure of that.

An alternative approach would be logging the user out after they submit the request to change their login.  Then the user would be expected to already be logged out.  Or a variant on that, clear the session if the user is already logged in if they go to the verify login change page.  However, both of those approaches present some minor usability issues.

I'll try to make a decision on this later today.  Likely, I'll just merge this if I can't find any security issues with it.

jeremyevans:
After reviewing the code, it definitely expects the user to not already be logged in, which is why there is support for autologin.  I agree the current behavior that requires logout before verifying a login change is suboptimal.  Therefore I think the best approach is to accept this, but if the user is already logged in, clear the session (equivalent of logout) before the route.  If autologin is enabled, the session is cleared on a succesful request anyway, and if it isn't enabled, the user won't be logged in, and logging in will clear the session.

jeremyevans:
Upon further review, the default behavior in this case is to allow this to be used while logged in.  So this only changes things if `already_logged_in` is used.  I guess it makes sense to consider this as a page that makes sense to access both when logged in and when not logged in, and therefore calling `already_logged_in` seems unnecessary.  So I don't think we need to clear the session for this, it should be acceptable as-is.

PR #89, opened 2020-05-04
Generate unlock account key outside of #send_unlock_account_email

janko:
This makes it easier to override `#send_unlock_account_email` with custom email sending, as it means that calling `#unlock_account_email_link` will just work, just like with other `#send_*_email` methods (see [rodauth-rails recommendation](https://github.com/janko/rodauth-rails/blob/7562530ea78cbf7a8bbffa8208915a0623473c1a/lib/generators/rodauth/templates/lib/rodauth_app.rb#L44-L61)).

Previously trying to call `#unlock_account_email_link` inside overridden `#send_unlock_account_email` would return an email link with empty token value, because the `@unlock_account_key_value` variable wasn't being set.

### DIFF
--- a/lib/rodauth/features/lockout.rb
+++ b/lib/rodauth/features/lockout.rb
@@ -73,6 +73,7 @@ module Rodauth
             redirect unlock_account_email_recently_sent_redirect
           end
 
+          @unlock_account_key_value = get_unlock_account_key
           transaction do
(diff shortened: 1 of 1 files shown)


jeremyevans:
Thanks for the patch!  I'll test and merge this later today.

PR #91, opened 2020-05-09
Don't consider account open if in unverified grace period without password

janko:
When the account is in the unverified grace period, but its password has not yet been set (i.e. `verify_account_set_password?` was set to true, delaying setting password until account verification), we should ensure that we prevent logins for this account.

I originally thought it's possible to log in to such account without password, but it seems that login already won't work in this case, because `#password_match?` method requires that the account has set password. However, in this case the user will get an "invalid password" error message, which is not exactly accurate.

I think it's safer to explicitly handle this scenario and not consider the account open in this case, which returns the correct error message that the account is awaiting verification. Essentially, if you're using verify_account_grace_period and delaying password set (i.e. setting `verify_account-set_password?` to true), you can only use the account unverified until the session expires, which is reasonable to me.

### DIFF
--- a/lib/rodauth/features/verify_account_grace_period.rb
+++ b/lib/rodauth/features/verify_account_grace_period.rb
@@ -23,7 +23,7 @@ module Rodauth
     end
 
     def open_account?
-      super || account_in_unverified_grace_period?
+      super || (account_in_unverified_grace_period? && has_password?)
     end
 
     def verify_account_set_password?
--- a/spec/verify_account_grace_period_spec.rb
+++ b/spec/verify_account_grace_period_spec.rb
@@ -140,6 +140,28 @@ describe 'Rodauth verify_account_grace_period feature' do
     page.body.must_include('Logged Intrue')
   end
 
+  it "should not allow login when password was not set" do
+    rodauth do
+      enable :login, :logout, :verify_account_grace_period
+      verify_account_set_password? true
+    end
+    roda do |r|
+      r.rodauth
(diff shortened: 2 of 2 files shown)


jeremyevans:
Thanks for the patch.

`verify_account_set_password?` is set to false by default in the verify_account_grace_period feature, so this would have to be turned on explicitly, and it doesn't make sense to turn it on for the reason you mentioned.  So this would only help cases where Rodauth has been configured in a strange way.

If I'm understanding this correctly, the only difference in behavior after this patch is the error message received.  Before the patch, it's an invalid password error.  After the patch, it's an unverified account error.  I'm not yet convinced that this change in error message is an improvement, but I'm not necessarily against it.

If you are using the change_password feature, change your password in the same session after the create account autologin, then you can logout and then login with the new password inside the grace period before verifying, correct?  To me, this indicates that this is a password issue and not an unverified account issue.

Is there an angle I'm not considering?

janko:
Thanks for your reply.

The reason I wanted to delay setting password until account verification (i.e. set `verify_account_set_password?` to false) is to decrease the amount of information that needs to be entered in order to create an account.

In my case, I want to use both this setting and `verify_account_grace_period`, because in terms of UX I though it doesn't make much sense delaying setting the password if the user cannot use their account before they verify it.

Initially I thought that if the password wasn't set, that Rodauth would allow logging in without password during the unverified grace period. But I then found out that it's not the case. I now realize that this behaviour is probably there for other reasons as well, and I agree that `invalid password` is fine for the error message.

My only goal with this PR was to ensure that login is disallowed in this case, but since it's already the case, I will close the issue 👍

PR #96, opened 2020-05-24
Move checking CSRF token into #check_csrf method

janko:
This allows the CSRF checking logic to be overridden. For example, rodauth-rails gem can now override this method to use Action Controller for checking CSRF tokens. This means when the user or a 3rd-party gem wants to disable CSRF protection for certain routes, they can have `#check_csrf?` return false for those routes, and that would work with rodauth-rails as well (currently rodauth-rails overrides `#before_rodauth` to do its CSRF protection).

cc @HoneyryderChuck


jeremyevans:
Thanks for the patch! This looks good to me, I'll test and merge on Tuesday.

PR #97, opened 2020-05-27
Use JWT json_response_body before halt to allow customize JSON response

arthurmmoreira:


jeremyevans:
Thanks for the patch!  This looks correct, I'll merge and test later today.

PR #98, opened 2020-06-03
Use 'new-password' on create-account template

nicolasmlv:
On /create-account, the password-field should be `autocomplete='new-password'` in order for browser to suggest new secure password

Screenshot 2020-06-03 at 03 39 50

Unfortunately there is some copy/paste from password-field.str, so I am not sure it is the best implementation.

Not sure if there should be a new `new-password-field.str`... Or if it is not a required feature and in this case feel free to cancel the PR 👍

In my specific case, I use https://github.com/janko/rodauth-rails and I overrided the views in order to send a parameter to `<%= render "password_field", create_account: true %>` and then use `local_assigns` to choose between `new-password` or `current-password`

### DIFF
--- a/templates/create-account.str
+++ b/templates/create-account.str
@@ -3,7 +3,12 @@
   #{rodauth.csrf_tag}
   #{rodauth.render('login-field')}
   #{rodauth.render('login-confirm-field') if rodauth.require_login_confirmation?}
-  #{rodauth.render('password-field') if rodauth.create_account_set_password?}
+  #{if rodauth.create_account_set_password?}
+    <div class="form-group">
+      <label for="password">#{rodauth.password_label}#{rodauth.input_field_label_suffix}</label>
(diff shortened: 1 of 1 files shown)


jeremyevans:
I agree with changing the autocomplete to new-password in this case.  As to the implementation, did you test your approach at all?  I didn't, but looking at it I would be surprised if it worked (str engine is ruby string interpolation, not like erb).

To handle this issue with the minimum of problems, we could add `password_field_autocomplete` accessor, set it to `new-password` in the create_account route,  and in password-field.str, use `:autocomplete=>rodauth.password_field_autocomplete || 'current-password'`.

nicolasmlv:
No I didn't test, i am not familiar with rodauth repo and the templating system, maybe I should've just created an issue 😄

I don't know where we should add `password_field_autocomplete` and where set it to true

PR #99, opened 2020-06-07
Don't check CSRF token on JSON requests

janko:
JSON requests should probably not be expected to provide a CSRF token, because CSRF protection is normally not part of a JSON API. This makes JSON requests in Rodauth with both HTML and JWT route versions enabled (i.e. with `json: true`) work automatically, the user doesn't need to disable CSRF protection for JSON routes.

/cc @nicolas-besnard

### DIFF
--- a/lib/rodauth/features/jwt.rb
+++ b/lib/rodauth/features/jwt.rb
@@ -140,6 +140,11 @@ module Rodauth
 
     private
 
+    def check_csrf?
+      return false if json_request?
+      super
+    end
+
     def before_rodauth
       if json_request?
         if jwt_check_accept? && (accept = request.env['HTTP_ACCEPT']) && accept !~ json_accept_regexp
--- a/spec/jwt_spec.rb
+++ b/spec/jwt_spec.rb
@@ -106,6 +106,21 @@ describe 'Rodauth login feature' do
     page.find('#notice_flash').text.must_equal 'You have been logged in'
   end
 
+  it "should not check CSRF for json requests" do
+    rodauth do
+      enable :login, :jwt
+      jwt_secret '1'
+      only_json? false
+    end
+    roda(:jwt_html) do |r|
+      r.rodauth
+      view(:content=>'1')
+    end
+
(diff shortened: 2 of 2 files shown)


jeremyevans:
Thanks for the this patch.  This makes sense to me, I'll merge it tomorrow.

HoneyryderChuck:
Wanted to raise: should we equate jwt authentication with json payloads? I mean, I could have an API using sessions. Shouldn't this all be part of base.rb?

jeremyevans:
By default, Rodauth treats JSON requests to Rodauth routes as using the JSON API.  You can override `check_csrf?` (which is in base.rb) if you want different behavior, this is just changing the default if the jwt feature is loaded.  I think this new default makes more sense, since most JSON APIs are not going to want CSRF protection.

It would probably be best to change this patch from using `json_request?` to using `use_jwt?`, as that is more similar to how the rest of the jwt feature works.

HoneyryderChuck:
> By default, Rodauth treats JSON requests to Rodauth routes as using the JSON API.

So at least "json_request?" should be a base.rb method?

jeremyevans:
I can't currently think of a reason that `json_request?` should be a base method, if nothing outside of the jwt feature uses it.  Am I missing something?

HoneyryderChuck:
[I'm defining it (when not available) in rodauth-oauth](https://gitlab.com/honeyryderchuck/rodauth-oauth/-/blob/master/lib/rodauth/features/oauth.rb#L215-222). The reasoning is mostly for determining what the mime-type/response format will be, which IMO shouldn't be dependent of using JWT (I can use API with sessions). I'm also using it to decide whether to check the csrf or not, which is what this PR is doing.

jeremyevans:
I understand where you are coming from, and see how adding `json_request?` would make rodauth-oauth implementation easier. However, with no internal uses, the fact that some external Rodauth features may want to check for JSON is not sufficient reason to move `json_request?` to base.

HoneyryderChuck:
@jeremyevans fair enough, I couldn't remove my code anyway to keep support for older versions of rodauth.

PR #102, opened 2020-06-14
WIP - Add disposable-provider plugin

nicolas-besnard:
* Block registration if account creation is using a disposable provider


jeremyevans:
I think the idea for this is fine, but it is best shipped as an external gem.  For one, I don't want to maintain a list of disposable providers.  If you setup an external gem for this, please submit a pull request to add a link to it to the website.

PR #104, opened 2020-06-29
Add "How-To" articles for common scenarios

janko:
Currently Rodauth feature documentation is mainly a reference of all configuration methods, which is good for people already familiar with Rodauth, but can be challenging for people just starting out.

So, as we discussed, I wrote some "How-To" articles for various use cases that are commonly found, which should hopefully help people achieve what they want more easily. These guides are inspired by the ["How-To" articles in the Devise wiki](https://github.com/heartcombo/devise/wiki/How-Tos).

They're in a format of short and focused tips, most of which I haven't actually tested, but figured they should work from my knowledge of Rodauth's source code. Since they're of different nature than the internals guide, I thought it would make sense to put them in a separate section.

### DIFF
--- a/doc/howto/admin_activation.rdoc
+++ b/doc/howto/admin_activation.rdoc
@@ -1,0 +1,31 @@
+= How To: Require admin account verification
+
+There are scenarios in which, instead of allowing the user to confirm their
+own created account, you may want to have an admin or moderator activate new
+accounts. This can be achieved by sending the account verification email to the
+admin:
+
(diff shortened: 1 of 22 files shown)


jeremyevans:
Awesome, thank you so much for working on this!  I'll merge this later today.

jeremyevans:
Merged at: 028f4765ba5ad1da4a3a46f04adbb3ece0c63526

PR #107, opened 2020-07-18
Make validating login email more configurable

janko:
Most authentication frameworks allow overriding the email regex, so I thought we could add that setting here as well:

```rb
# uses the email_validator gem
login_email_regexp EmailValidator.regexp
```

I think it's also useful to allow overriding the method that validates the email, for when we want more complex validation than just using a regex:

```rb
# uses the truemail gem
login_valid_email? { |login| Truemail.valid?(login) }
```

Finally, I also extracted the validation error for when login is not a valid email address:

```rb
login_not_valid_email_message "not a valid email address"
```


jeremyevans:
Thanks for the patch! These changes all look good.  I've already done the final testing for 2.2.0, which will be released on Monday, so I'll test and merge this after the release.

jeremyevans:
Merged at a2550ac62ee6362fc37552dd529a764fd86c1f0e

PR #108, opened 2020-07-20
Update anchor link for "Guides" heading

janko:
The original heading text was "How-To", which was then changed to "Guides". This updates the anchor identifier accordingly as well.


jeremyevans:
Thanks for the patch!

PR #110, opened 2020-07-23
Add route before hook for jwt-refresh

AlexeyMatskevich:
I noticed that the documentation describes the hook http://rodauth.jeremyevans.net/rdoc/files/doc/jwt_refresh_rdoc.html but it doesn't work and it looks like it's not written in the code


jeremyevans:
Thanks for the patch!

PR #112, opened 2020-07-25
Add ability to use catch_error behavior into before and after hooks

AlexeyMatskevich:
Many actions allow to use behavior of catch_error with hooks.
This makes it easy to return your errors with throw_error_status.

```ruby
before_close_account do
  throw_error_status(field, filed_param, error_message) if some_code
end
```

I think it would be convenient to have the same opportunity in this action.

Are there any reasons not to use this approach everywhere?
I could contribute on the remaining actions.

Thanks, Alexey

### DIFF
--- a/lib/rodauth/features/close_account.rb
+++ b/lib/rodauth/features/close_account.rb
@@ -33,7 +33,11 @@ module Rodauth
       end
 
       r.post do
-        if !close_account_requires_password? || password_match?(param(password_param))
+        catch_error do
+          if close_account_requires_password? && !password_match?(param(password_param))
+            throw_error_status(invalid_password_error_status, password_param, invalid_password_message)
+          end
+
           transaction do
             before_close_account
             close_account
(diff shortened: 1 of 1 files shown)


jeremyevans:
I suppose it's fine to change the implementation to use `catch_error`.  Currently, I've only used in places where it is useful internally, but extending that to places where it could be useful in external code could make things more flexible.

One issue with this is that `throw_error_status` is a private API.  The fact that Rodauth uses catch/throw internally is an implementation detail, it's not something that people using Rodauth should be relying on.  However, if people are already using `throw_error_status` in their own configuration methods, maybe the ship has already sailed in that regard.  I don't expect to change the internals in this regard in the future, so it's probably OK.

I'll test this in a couple days, and assuming no problems I'll merge it.

AlexeyMatskevich:
It would be convenient to have a code convention for returning errors and interrupting code execution.

jeremyevans:
Cherry-picked as 50440b69d1c4191893dd824e47d049cf26ed6398

PR #113, opened 2020-08-04
Remove unnecessary variable assignment in jwt_refresh

janko:
This isn't needed anymore after https://github.com/jeremyevans/rodauth/commit/4c52564817bea2a982aa8fb7276925a9d345d66f.


jeremyevans:
Thanks for the patch!

PR #114, opened 2020-08-10
Rename #_login to #login and make it public

janko:
A `#login` method is useful when having custom authentication logic, such as using OmniAuth to allow people to authenticate through 3rd-party services, for example:

```rb
r.get "/auth/callback" do
  auth  = request.env["omniauth.auth"]
  email = auth["info"]["email"]

  rodauth.account_from_login(email)
  rodauth.login('omniauth')
end
```

Here the user is not logging in via email/password, but we still want to reuse things like login hooks, notice flash and redirect logic.

I would like to create an OmniAuth integration at some point, which would be able to use Rodauth's private `#_login` method, but I would still like the rely on the method being here to stay.

### DIFF
--- a/doc/login.rdoc
+++ b/doc/login.rdoc
@@ -3,6 +3,13 @@
 The login feature implements a login page.  It's the most commonly
 used feature.
 
+In addition to the auth methods below, it provides a +login+ method that wraps
++login_session+, running login hooks and redirecting to the configured
+location.
+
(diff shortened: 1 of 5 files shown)


jeremyevans:
This looks good to me and I'm OK with making `login` public.  However, please add a test for the usage of `_login`.

janko:
@jeremyevans Added a test for `#_login` ✅

PR #115, opened 2020-08-10
Move session expiration redirect into a macro

janko:
Since this redirect doesn't have any advanced logic (such as `confirm_password_redirect`), we can define it using the `redirect` macro, for consistency with how we're defining other redirects. For me this is provides a slight readability improvement, but feel free to close if you feel the change is too minor.


jeremyevans:
I think this makes sense.  Thanks for the patch!

jeremyevans:
Merged as acada0f32230bd25e131486bd374faa04b2cefd7

PR #116, opened 2020-08-13
Simplify multi-phase login logic

janko:
I came up with a way to simplify the multi-phase login code in the login endpoint. I like that now only the `using_multi_phase_login?` branch is concerned with multi-phase login logic, the rest of the code is handling regular login.

I've also extracted the `#halt` method in the process, which I found useful in other places as well.


janko:
Oh, I missed an edge case, caught by the tests (entering a valid login and a password), nevermind 😛

PR #117, opened 2020-08-16
Add support for Postgres UUID primary keys

monorkin:
@janko and I have been working on adding Postgres UUID support to Rodauth.

This patch implements the needed changes to support UUIDs as primary and foreign keys where appropriate.
The user implementing Rodauth in their system has to use UUIDs as primary keys when creating their tables or tables that rodauth internally uses. A full example is available in `spec/migrate/001_tables.rb`.

Comments:
* The `account_previous_password_hashes` uses a Bignum as the primary key as the row ID is never referenced except for sorting. UUID could be used here, but it would require the introduction of a new column to make the sort stable, and existing systems using Rodauth would require a migration task.
* The `account_statuses` uses Bignum as the primary key as the number of elements is expected to be low and it has to be predictable
* The audit log tests now search for an entry with the message they expect as sorting based on timestamps (that is the `CURRENT_TIMESTAMP` and `NOW` functions) returns the time the transaction initiated, and therefore all rows created within the test (as they share the same transaction) are created with an identical timestamp.
* Some lines ending with `to_s` in the test suite have been changed to end with `inspect` as to differentiate results containing UUIDs form results containing numbers (`'foo'.to_s # => "foo"` while `'foo'.inspect # => "\"foo\""`).
* We have added a new entry in the Travis matrix to test UUID support
* At some places we read a table's primary key type so that the user doesn't have to specify the type of primary key they use.

Co-authored-by: Janko Marohnić <janko.marohnic@gmail.com>

### DIFF
--- a/spec/migrate/001_tables.rb
+++ b/spec/migrate/001_tables.rb
@@ -1,10 +1,6 @@
 Sequel.migration do
   up do
-    primary_key_type = if ENV['RODAUTH_SPEC_UUID'] && database_type == :postgres
-                         :uuid
-                       else
-                         :bigint
-                       end
+    primary_key_type = ENV['RODAUTH_SPEC_UUID'] && database_type == :postgres ? :uuid : :bigint
 
     extension :date_arithmetic
 
--- a/spec/migrate/001_tables.rb
+++ b/spec/migrate/001_tables.rb
@@ -1,5 +1,11 @@
 Sequel.migration do
   up do
+    primary_key_type = if ENV['RODAUTH_SPEC_UUID'] && database_type == :postgres
+                         :uuid
+                       else
+                         :bigint
+                       end
+
     extension :date_arithmetic
 
     # Used by the account verification and close account features
@@ -11,7 +17,11 @@ Sequel.migration do
 
     db = self
     create_table(:accounts) do
-      primary_key :id, :type=>:Bignum
+      if primary_key_type == :uuid
+        uuid :id, :primary_key=>true, :default=>Sequel.function(:gen_random_uuid)
+      else
+        primary_key :id, :type=>:Bignum
+      end
       foreign_key :status_id, :account_statuses, :null=>false, :default=>1
       if db.database_type == :postgres
         citext :email, :null=>false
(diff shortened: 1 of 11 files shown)


spec/migrate_password/001_tables.rb
```diff
@@ -47,7 +53,7 @@
       run "REVOKE ALL ON FUNCTION rodauth_previous_password_hash_match(int8, text) FROM public"
       run "GRANT INSERT, UPDATE, DELETE ON account_previous_password_hashes TO #{user}"
       run "GRANT SELECT(id, account_id) ON account_previous_password_hashes TO #{user}"
-      run "GRANT USAGE ON account_previous_password_hashes_id_seq TO #{user}"
+      # run "GRANT USAGE ON account_previous_password_hashes_id_seq TO #{user}"
```
jeremyevans, line 56: It seems odd to just comment this out.  Shouldn't it be conditional (run for Bigint, not run for uuid)?

spec/migrate/001_tables.rb
```diff
@@ -1,5 +1,11 @@
 Sequel.migration do
   up do
+    primary_key_type = if ENV['RODAUTH_SPEC_UUID'] && database_type == :postgres
```
jeremyevans, line 3: Style-wise, I would prefer the ternary operator on a single line in this case (and the other spec migration).

jeremyevans:
Thanks, the changes look good.  I'll try to test and merge tomorrow.

spec/migrate_password/001_tables.rb
monorkin: This was accidentally added to the patch while we were working out how to support UUIDs with that table.

In the final implementation `account_previous_password_hashes` always uses a `int8` primary key, therefore the password user should always be granted usage on `account_previous_password_hashes_id_seq`.

I've pushed a patch to uncomment the line.

spec/migrate/001_tables.rb
monorkin: I've updated the three files with this check and converted them to ternary operators on a single line.

jeremyevans:
Commits cherry-picked with a couple extra fixes at 72130a67e8411be46edc382247185c01cfe1c357.  In the future, please rebase the branch against master before submitting a pull request.  Thanks for your work on this!

PR #118, opened 2020-08-21
Add ability to assign a public key

AlexeyMatskevich:
Currently, there is no way to use the `RSA` algorithm due to the fact that different keys cannot be assigned for encoding and decoding.

### DIFF
--- a/lib/rodauth/features/jwt.rb
+++ b/lib/rodauth/features/jwt.rb
@@ -27,6 +27,7 @@ module Rodauth
     auth_value_methods(
       :only_json?,
       :jwt_secret,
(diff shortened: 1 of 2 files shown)


jeremyevans:
This seems like a OK idea, though I'm not sure why you would choose to use public key encryption when you control both the public and private keys.  Public key encryption generally only makes sense if the public key must be shared.

A couple suggestions in terms of implementation: 

* The spec doesn't appear to check that you can decode the JWT.  You should have it make a request after login that can determine whether the session is logged in.
* The doc folder needs to be updated to document the `jwt_public_secret` configuration method.

PR #121, opened 2020-09-10
Add session_key_prefix configuration for adding prefix to all session keys

janko:
In response to https://github.com/janko/rodauth-rails/issues/11, this PR adds the `session_key_prefix` configuration for prefixing all session keys. This is useful when using multiple Rodauth configurations, where it's usually desired to have the session state separated between configurations.

```rb
plugin :rodauth do
  # ...
end

plugin :rodauth, name: :admin do
  # ...
  session_key_prefix "admin_"
end
```

cc @bjeanes

### DIFF
--- a/README.rdoc
+++ b/README.rdoc
@@ -1064,6 +1064,15 @@ the name as an argument to use that configuration:
     r.rodauth
   end
 
+By default, alternate configurations will use the same session keys as the
+primary configuration, which may be undesirable. To ensure session state is
+separated between configurations, you can set a session key prefix for
+alternate configurations:
+
+  plugin :rodauth, :name=>:secondary do
+    session_key_prefix "secondary_"
+  end
+
 === With Password Hashes Inside the Accounts Table
 
 You can use Rodauth if you are storing password hashes in the same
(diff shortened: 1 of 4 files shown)


janko:
I'll also update the README with the suggestion to use `session_key_prefix`.

bjeanes:
🙏🙏

bjeanes:
I'm on phone so haven't checked diff closely, but will this cover cookies too? Things like remember cookie need to be different also

jeremyevans:
Thanks, this looks good, will merge shortly!

This would only affect session keys, it would not affect remember cookie names.  I don't think it should, based on the name, but it may be worth mentioning the remember cookie name in the documentation.  I'll add that after merging this.

bjeanes:
Is the remember cookie the only explicit cookie? I ended up doing something like (I'm on a different computer so this is from memory):

``` ruby
methods.grep(/cookie|session/).grep(/_key$/).each do |m|
  send(m) { "admin_#{super()}" }
end
```

jeremyevans:
I believe remember is the only feature that ships with Rodauth that uses a separate cookie.  Everything else uses the session, and the `session_key_prefix` can handle that.

PR #124, opened 2020-09-29
Make `new_password_label` translatable

bjeanes:
I believe this has likely just been overlooked. I noticed it when the label wasn't changing when tweaking locale texts!


jeremyevans:
You are correct, this was an oversight.  Thanks!

PR #127, opened 2020-11-02
Fix setup of subsequent Webauthn identities

bjeanes:
The `webauthn` feature is designed to accommodate multiple Webauthn identies being created against a single Rodauth account. However, presently, this is broken because Rodauth is not correctly preparing the `excludeCredentials` option on credential creation options.

This PR does two things:

* Unpacks and prepares the correct representation for the credential IDs to exclude.
* Extracts the repeated packing and unpacking between (what appears to be) url-safe Base64 and the array buffer than the Webauthn browser API expects. This is purely opportunistic but it seemed like a subtle place for a bug to creep into in the future. I'm happy to excise this from the PR if you don't want this change.

There is no test for this because the Rodauth tests don't exercise JS (and Webauthn requires TLS connections anyway) so I literally have no idea how I could add a test for this.

### DIFF
--- a/javascript/webauthn_setup.js
+++ b/javascript/webauthn_setup.js
@@ -9,6 +9,7 @@
       var opts = JSON.parse(element.getAttribute("data-credential-options"));
       opts.challenge = unpack(opts.challenge);
       opts.user.id = unpack(opts.user.id);
+      opts.excludeCredentials.forEach(function(cred) { cred.id = unpack(cred.id); });
       //console.log(opts);
       navigator.credentials.create({publicKey: opts}).
         then(function(cred){
--- a/javascript/webauthn_setup.js
+++ b/javascript/webauthn_setup.js
@@ -1,26 +1,28 @@
 (function() {
+  var pack = function(v) { return btoa(String.fromCharCode.apply(null, new Uint8Array(e))).replace(/\+/g, '-').replace(/\//g, '_').replace(/=/g, ''); };
+  var unpack = function(v) { return Uint8Array.from(atob(v.replace(/-/g, '+').replace(/_/g, '/')), c => c.charCodeAt(0)); };
   var element = document.getElementById('webauthn-setup-form');
   var f = function(e) {
     //console.log(e);
     e.preventDefault();
     if (navigator.credentials) {
       var opts = JSON.parse(element.getAttribute("data-credential-options"));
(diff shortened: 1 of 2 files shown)


javascript/webauthn_setup.js
```diff
@@ -1,26 +1,29 @@
...
   var f = function(e) {
     //console.log(e);
     e.preventDefault();
     if (navigator.credentials) {
       var opts = JSON.parse(element.getAttribute("data-credential-options"));
-      opts.challenge = Uint8Array.from(atob(opts.challenge.replace(/-/g, '+').replace(/_/g, '/')), c => c.charCodeAt(0));
-      opts.user.id = Uint8Array.from(atob(opts.user.id.replace(/-/g, '+').replace(/_/g, '/')), c => c.charCodeAt(0));
+      opts.challenge = unpack(opts.challenge);
+      opts.user.id = unpack(opts.user.id);
+      opts.excludeCredentials.forEach(function(cred) { cred.id = unpack(cred.id); });
```
bjeanes, line 12: This is the singular line which is not merely an extraction of the `atob`/`btoa` dance.

I suggest viewing the two commits in isolation for review purposes.

jeremyevans:
Thanks for the patch!  I remember when I was working on this code originally, I made sure you couldn't register the same key multiple times.  Looks like that technically worked, but I never actually tried registering a second key, so I didn't realize there was a bug in the javascript preventing registering a separate key.

There appears to be one bug in the resulting javascript after the change (`e` instead of `v`), but I've fixed that locally.

bjeanes:
Ah thanks for that. I thought the risk was there since I don't have a TLS set up here and it was non-trivial to test. I was hoping the purely mechanical nature of this refactor would have my back :|

jeremyevans:
Thankfully, I have a setup with a local nginx->unicorn with the demo site allowing webauthn testing, so it was possible for me to manually test the javascript changes and ensure everything still worked.  This is something I probably should have checked earlier, but I only have a single key so it wasn't easy.  I had to manually modify the database table to change the webauthn_id after registering the key and then try registering it again to check that it did work.

bjeanes:
Makes sense.

I discovered this by trying to register my fingerprint as a second factor. So, if you ever need to try multiple keys again, try your existing (presumably FIDO/Yubikey) and then use either Chrome on Android or Safari on iOS or Safari on macOS to use try a bio-based second factor, presuming you have one of those three devices at your disposal.

PR #128, opened 2020-11-03
Add Osso to open source applications

sbauch:
The description is what the OAuth.net maintainers suggested when we added ourselves over there, but open to your edits - https://oauth.net/code/


jeremyevans:
Thanks for the patch!

PR #129, opened 2020-11-09
`around_rodauth`

bjeanes:
This PR adds an `around_rodauth` extension point for features to use. It allows a feature (such as the one provided by `rodauth-rails`) to conditionally allow the request to proceed (e.g. if a Rails-owned `before_action` has rendered or redirected).

I think the final design here is as you indicated on the mailing list @jeremyevans, but `&block` needs to be explicitly passed to `super()` because implicit `super` args are not available in the dynamic context.

cc @janko

### DIFF
--- a/doc/base.rdoc
+++ b/doc/base.rdoc
@@ -87,7 +87,9 @@ account_id :: The primary key value of the current account.
 account_session_value :: The primary value of the current account to store in the session when logging in.
 after_login :: Run arbitrary code after a successful login.
 after_login_failure :: Run arbitrary code after a login failure due to an invalid password.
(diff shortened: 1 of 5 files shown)


jeremyevans:
Thanks for the patch!  It looks mostly good, I think I can fix the remaining issues myself.  Will test and merge shortly.

jeremyevans:
Squash merged at 8376262094944dab8a671c3db118d36083beaba6

bjeanes:
Ah yes. I see I left some vestigial traces of my first, more generic, attempt there. Sorry about that!

PR #131, opened 2020-11-12
Avoid features being loaded multiple times

janko:
When loading a tree of dependencies, we need to immediately write down when we've loaded the dependency, otherwise in this case they can be loaded multiple times. I've added a spec that would return duplicate features with current code.

It doesn't cause any actual problems, but users of `rodauth.features` and `rodauth.routes` might rely that every feature is listed exactly once. In my case, I'm using these methods to list available routes.

### DIFF
--- a/lib/rodauth.rb
+++ b/lib/rodauth.rb
@@ -290,9 +290,11 @@ module Rodauth
     end
 
     def enable(*features)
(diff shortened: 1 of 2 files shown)


jeremyevans:
Thanks for the patch!  This definitely looks like a better way of doing it.

PR #132, opened 2020-11-15
Improve routing syntax consistency in confirm_password feature

janko:
This PR does a minor update to the routing syntax in confirm_password feature to make it consistent with other Rodauth features.


jeremyevans:
Thanks for the patch!

PR #134, opened 2020-11-27
Default remember cookie path to "/"

janko:
Cookies without a path behave unexpected on requests with multiple path components. When a `/a/b/c` request sets a cookie, that cookie appears to be stored for `/a/b` path. This cookie is visible in the inspector on Firefox, but not on Chrome. Attempting to delete the cookie doesn't work, because a subsequent request to `/a/b` still sends the cookie (but not a request to `/` or any other route). See [this StackOverFlow answer](https://stackoverflow.com/a/64515458/988414) for more details.

This can cause unexpected behaviour when using `#load_memory` before each request, as the user that has manually signed out can be autologged back in from the remember cookie if a request is made to a certain path.

Setting cookie path to `/` resolves this issue. Rails does this by default for its cookies as well.

### DIFF
--- a/doc/remember.rdoc
+++ b/doc/remember.rdoc
@@ -35,7 +35,7 @@ raw_remember_token_deadline :: A deadline before which to allow a raw remember t
 remember_additional_form_tags :: HTML fragment containing additional form tags to use on the change remember setting form.
 remember_button :: The text to use for the change remember settings button.
 remember_cookie_key :: The cookie name to use for the remember token.
-remember_cookie_options :: Any options to set for the remember cookie.
+remember_cookie_options :: Any options to set for the remember cookie. By default, the `:path` cookie option is set to `/`.
 remember_deadline_column :: The column name in the +remember_table+ storing the deadline after which the token will be ignored.
 remember_deadline_interval :: The amount of time for which to remember accounts, 14 days by default.  Only used if +set_deadline_values?+ is true.
 remember_disable_label :: The label for disabling remembering.
--- a/lib/rodauth/features/remember.rb
+++ b/lib/rodauth/features/remember.rb
@@ -132,11 +132,14 @@ module Rodauth
       opts = Hash[remember_cookie_options]
       opts[:value] = "#{account_id}_#{convert_token_key(remember_key_value)}"
       opts[:expires] = convert_timestamp(active_remember_key_ds.get(remember_deadline_column))
+      opts[:path] = "/" unless opts.key?(:path)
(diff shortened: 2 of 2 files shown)


jeremyevans:
I think this would affect backwards compatibility.  If I have two apps using Rodauth (or two Rodauth configurations in the same app), one mounted at `/a` and another at `/b`, using the default Rodauth settings with the remember feature, currently that situation works correctly.  With this change, then it starts sending  remember cookies for the `/b` app to the `/a` app and vice versa.

I agree with you that a cookie path of `/` is a better default, but I think that change should wait until Rodauth 3.   However, I'd like to hear your opinion on that before making a decision.

janko:
I agree that in these scenarios would introduce backwards incompatibility. However, I do think the tradeoff is worth it, given that this is not a common scenario, and that in most common scenarios it can cause unexpected behaviour as soon as the auth routes are nested. I would like to give two concrete examples:

1️⃣ If the Rodauth routes have a path prefix, e.g. `/auth`, and the account is automatically remembered after login (e.g. with `after_login { remember_login }`), then the remember feature won't work correctly. Because of the behaviour I described, when the remember cookie is added in `POST /auth/login` response, it will actually be added for the `/auth` path prefix. This means that when the session expires (e.g. browser is closed), visiting `/` will not autologin the user. The cookie will only be sent for `/auth*` paths, e.g. when visiting `/auth/login` again.

2️⃣ I recently added an [OmniAuth integration section](https://github.com/janko/rodauth-rails#omniauth) to the rodauth-rails README, which calls Rodauth's `#login` method in an `/auth/{provider}/callback` route. In this case I still had Rodauth routes on the top level, i.e. I didn't use `/auth` prefix for them. So, after the external login the user is redirected to the `/auth/{provider}/callback` route, which logs them in via `#login`, which also calls `#remember_login`, and the cookie is set to `/auth/{provider}`. When the user now logs out, the `forget_login` won't correctly delete the cookie. So when the user authenticates via 3rd party again, before my `/auth/{provider}/callback` handler was reached, `#load_memory` would autologin the account from remember cookie, even though the user explicitly signed out.

I think in 2️⃣ the reason why remember cookie wasn't being deleted on logout is because cookie paths are also taken into account when deleting cookies. So, if we have a cookie for `/auth/{provider}` path, then deleting the cookie in `POST /logout` response won't work, because it would only delete cookies with paths set one level down, i.e. `/`. This means I should probably update my PR to send the `:path` when removing cookies as well.

jeremyevans:
OK.  I guess I can accept the backwards compatibility breakage in this case.  Please update the documentation to reflect that the :path cookie option is set to `/` if not provided.

PR #135, opened 2020-11-29
Add :auto_remove_recovery_codes? option to RecoveryCodes feature

SilasSpet:
This adds an `auto_remove_recovery_codes?` option to the `recovery_codes` feature. It defaults to `false`. If set to `true` it will delete all recovery codes once the last MFA method (SMS, OTP or WebAuthn) has been removed. It won't delete recovery codes if an MFA method is removed and others are still active (e.g. disable OTP with SMS still being active).


jeremyevans:
This looks OK, but I think you need to call `super if defined?(super)` in your after hooks, otherwise if recovery_codes is loaded after other MFA features, it wouldn't call the user-defined hooked correctly.

I'll try to test and merge tomorrow.

SilasSpet:
Just added them!

PR #136, opened 2020-11-29
Allow Rodauth features to be in custom location

janko:
This allows manually loading Rodauth features before enabling them, allowing them to be in a location other than `rodauth/features/<feature>.rb` in the load path.

I would find this useful in rodauth-rails, where I have everything nested in `rodauth/rails/*`, including the `rails` feature which is at `rodauth/rails/feature.rb`. At the moment I also have `lib/rodauth/features/rails.rb`, which just requires that file, but it would be nice to be able to remove it.

### DIFF
--- a/lib/rodauth.rb
+++ b/lib/rodauth.rb
@@ -300,7 +300,7 @@ module Rodauth
     private
 
     def load_feature(feature_name)
-      require "rodauth/features/#{feature_name}"
+      require "rodauth/features/#{feature_name}" unless FEATURES[feature_name]
       feature = FEATURES[feature_name]
       enable(*feature.dependencies)
       extend feature.configuration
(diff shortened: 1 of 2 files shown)


jeremyevans:
This should be fine, but can you please add a test for it?

janko:
Done 👍 At first I thought it would be difficult to write a test, since it involves `require`s, but it turned out to be really simple.

jeremyevans:
Cherry-picked at 7a0a5403680ca4409d4bc221304f3195e5c85eed

bjeanes:
With the caveat that I haven't yet tested this, I think this will prevent reloading features (in Rails autoloader context) which I rely upon for a lot of local features that I use/refine in my Rails app.

I will test this to be sure and report back (likely this week). I'm just back from vacation and catching up on rodauth/rodauth-rails changes.

jeremyevans:
We can definitely add an option to disable the behavior if it is problematic.

bjeanes:
Maybe, but I clicked into this issue because I've wanted to be able to have my features anywhere on disk too. The ideal ergonomics here would be to allow Rodauth to skip caching the feature _entirely_ and let users `enable SomeFeatureModule`, but at that point it starts looking an awful lot like `include SomeFeatureModule` anyway...

PR #140, opened 2020-12-23
Fix #clear_session not clearing JWT session with sessions plugin

janko:
When sessions Roda plugin is used, `#clear_session` will call `scope.clear_session`, which unlike `session.clear` will not clear the `@session` hash.

So in JWT feature we make sure to call `session.clear` in `#clear_session` in case it wasn't called in the super method. I wasn't sure whether to add an `if scope.respond_to?(:clear_session)` here, I thought it probably wasn't worth repeating the conditional.

I also added assertions for setting empty JWT token when `#clear_session` is called but Rodauth isn't the one returning the response to the client. As I mentioned on the google group, this spec was actually the main purpose of my PR 😛 But then I discovered this potential issue through reading, so I thought I'd include a fix.

### DIFF
--- a/lib/rodauth/features/jwt.rb
+++ b/lib/rodauth/features/jwt.rb
@@ -73,7 +73,10 @@ module Rodauth
 
     def clear_session
       super
-      set_jwt if use_jwt?
+      if use_jwt?
+        session.clear
+        set_jwt
+      end
     end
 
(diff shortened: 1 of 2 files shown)


jeremyevans:
Good find!  I will test and merge shortly.

PR #141, opened 2020-12-28
Use jwt_refresh_token_key in #update_session override

janko:
I noticed this while reading the code, and I think `jwt_refresh_token_key` was meant to be used in `#update_session` override instead of the `'refresh_token'` string.

I've also noticed a few methods could be reused, so I included these changes in the PR as well.


jeremyevans:
Thanks for these fixes!

jeremyevans:
Cherry-picked at ee70ca827219e7ee38790af958e1e7dce825ec2b

PR #142, opened 2021-01-06
Forbid JavaScript from accessing remember cookie

janko:
Currently the remember cookie can be retrieved via `document.cookie`, which opens the surface for attacks if the app using Rodauth has a XSS vulnerability, as it allowing malicious JavaScript code to steal the remember token and the attacker to login to the user's account.

We mitigate this attack by setting `HttpOnly` for the remember cookie, which forbids JavaScript from accessing the cookie. This should not break code for most apps, because the client generally doesn't need to read this value, it only needs to send the cookie to the server (which happens automatically).

### DIFF
--- a/doc/remember.rdoc
+++ b/doc/remember.rdoc
@@ -35,7 +35,7 @@ raw_remember_token_deadline :: A deadline before which to allow a raw remember t
 remember_additional_form_tags :: HTML fragment containing additional form tags to use on the change remember setting form.
 remember_button :: The text to use for the change remember settings button.
 remember_cookie_key :: The cookie name to use for the remember token.
-remember_cookie_options :: Any options to set for the remember cookie. By default, the `:path` cookie option is set to `/`.
+remember_cookie_options :: Any options to set for the remember cookie. By default, the `:path` cookie option is set to `/`, and `:httponly` is set to `true`.
 remember_deadline_column :: The column name in the +remember_table+ storing the deadline after which the token will be ignored.
 remember_deadline_interval :: The amount of time for which to remember accounts, 14 days by default.  Only used if +set_deadline_values?+ is true.
 remember_disable_label :: The label for disabling remembering.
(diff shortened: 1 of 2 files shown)


jeremyevans:
I agree that this is a better default.  This change may break some configurations, but those can be handled by setting `remember_cookie_options(:httponly=>false)`.  While this isn't necessarily a vulnerability in Rodauth itself, as it is security related, I'll put out a new release with it later today.

jeremyevans:
Cherry-picked at 551938d931b231904102624e2ca406c78ffd4138

HoneyryderChuck:
Don't these fall into the category of settings that belong in the rack middleware? Should rodauth set httponly prop if I set it to false app-wide?:

```ruby
use Rack::Session::Cookie, {:httponly => true, :secure => true }
```

jeremyevans:
This is the remember cookie, not the session cookie.  Rodauth doesn't do anything with the session, since it's possible the session won't even use cookies.

janko:
@HoneyryderChuck I happen to have been studying cookies & session recently, and session cookie store is not actually related to setting cookies manually. The `Rack::Session::Cookie` middleware enables storing arbitrary data into a single signed cookie (e.g. `_your_app_session`).

The reason the remember feature sets cookies manually is partly because it needs to set its own expiration that's independent of the cookie that holds session data. And the cookie that holds the session data is usually a "session cookie", which is a cookie for which expiration wasn't set, and it expires when the browser session is closed.

You probably already knew most or all of these things, but reading the Rails source code the past 2 days really helped me to better understand the relationship between sessions and cookies.

At first I thought it would useful to automatically apply session options for cookies; I've been thinking of having rodauth-rails apply Rails' `session_options` to Rodauth cookies. But then I realized that anyone setting cookies manually (e.g. in Rails via `cookies`) already needs to take care of applying the cookies options manually, as nothing is inherited from session options in Rails. If this was fine for Rails, then I figured it's fine for Rodauth too. It's worth noting that [Devise inherits session options for remember feature](https://github.com/heartcombo/devise/blob/98fc5e8e396b66b826528811287ea6680a6d0757/lib/devise/controllers/rememberable.rb#L11), though Sorcery doesn't.

janko:
@jeremyevans Thank you for getting it released so quickly 🤘 (and for adding tests, that makes total sense, especially considering that a naive implementation using `||=` wouldn't allow disabling `:httponly`)

HoneyryderChuck:
You're both right, I got this mixed up.

PR #144, opened 2021-01-10
Make remember cookie secure by default with SSL

janko:
Secure cookies are only sent to the server when a request is made with the HTTPS scheme, and therefore are more resistant to man-in-the-middle attacks, [according to MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie). Given this, I thought it might be safer make the remember cookie `secure` by default when the request is made over SSL.

I believe this is mostly a backwards compatible change, except for sites that are accessible both over `http://` and `https://`, where `http://` doesn't redirect to `https://`. However, this configuration is insecure by itself, so I thought it's fine.

I've also took the liberty of DRYing up remember feature tests a bit, and moving assertions for all cookie attributes together in a dedicated test.


jeremyevans:
Thanks for the patch!  I agree this is a good idea.

PR #145, opened 2021-01-16
Extract JSON feature from JWT

janko:
At work we're using custom token-based authentication over JSON API, and I would like to migrate us to Rodauth soon, so I started reading about JWT last week to learn more. After finding out about the challenges in making JWT authentication secure (e.g. storing the token in localstorage/cookies makes it retrievable through XSS, which then requires JWT expiration & refresh tokens), and given that we have a Vue frontend and a Rails app running on the same origin, I realized it would be easiest for us to use regular cookie-based authentication via rack session.

To make this possible, this pull request extracts JSON API functionality from the `jwt` feature into a new **`json`** feature. This was originally suggested in [this google group thread](https://groups.google.com/g/rodauth/c/9nfZPQvyW44/m/KRwM_j7NEAAJ) by @mwpastore, which also provides more details and some links on the rationale.

The extraction was very straightforward (for both application code, tests, and docs), it turned out these features were already nicely decoupled. I like how it's easier now to understand both features individually.

### Implementation

The `json` feature adds a new `use_json?` method, which is used for determining whether to return a JSON response. `use_jwt?` is still used for JWT-specific behaviour, such session management and skipping CSRF protection. This requires any request passing the JWT token to also set `Content-Type` request header to `application/json` (or whatever is in the `json_request_content_type_regexp`), but I think that was needed before anyway, otherwise Roda wouldn't correctly parse the request body.

I've also created a `_return_json_response` method to DRY up places which need to return the JSON response without setting the JWT token ( `#session`, `#before_rodauth` and the `/jwt-refresh` route). This change wasn't strictly needed, but I felt like DRYing up these places as I saw benefits in readability. The `return_json_response` method is then overridden by the `jwt` feature to also set the JWT token.

I've replaced `jwt_check_accept?` setting with `json_check_accept?`, and left the former as a deprecated alias.

For the JSON-only mode I wanted to recommend setting `only_json? true` instead of passing the `json: :only` plugin option , because it's expected this feature will be used with the rack session in the cookie, so I wanted to make sure the CSRF roda plugin is loaded.

### Tests 

I thought it would be good idea for each feature that's tested with `jwt` & JWT session to also be tested with `json` & rack session, so I've wrapped these tests into a loop. For the remember feature tests I needed to improve cookie handling accuracy a bit, because it now needed to manage two different cookies (`rack.session` and `_remember`).

One thing I need help with is resolving the `Unable to decode session: no data` warnings from Roda's `sessions` plugin, which occur after the session has been cleared on logout. I think the reason why this *isn't* happening with HTML requests is that there is always a flash message there keeping the session data non-empty and preventing the `sessions` plugin from deleting the cookie.


jeremyevans:
Thanks for the patch!  I agree decoupling JSON from JWT makes sense.  I'll review on Tuesday.

For the sessions warning, I think Roda assumes if the session cookie key exists, that the cookie value shouldn't be completely empty.  It should at the very least be an Base64 encoded value containing an empty JSON hash (after the sessions plugin decodes it).  If the sessions plugin is deleting the cookie, it shouldn't be sent in future requests.  Maybe rack-test still sends the cookie even though it should not?  The warning itself is harmless, we could certainly just ignore it in the Rodauth specs.

rmacklin (commented):
Thanks for working on this! I came here from your `rodauth-rails` project to see if this was supported since one of our apps also uses a JSON API with cookie-based authentication. Glad to see you were already working on it :) 

Regarding:
> I've replaced `jwt_check_accept?` setting with `json_check_accept?`, and left the former as a deprecated alias.

I noticed two spots in the documentation that are referencing the deprecated name - see below

janko:
@rmacklin Good catch, I've updated those places now to use `json_check_accept?`.

@jeremyevans Thanks for the leads, they helped me resolve the warnings. The tests were indeed sending an empty `rack.session` cookie, it was because on my computer the timezone of the expiration date for cookie deletion was `GMT` instead of `-0000`, so it wasn't being detected.

HoneyryderChuck:
Great initiative @janko! I had also thought about this before while doing rodauth-oauth, but never actually moved to see if it was possible to decouple both. Tha k you!

jeremyevans:
I'm doing some final testing on this now.  This took longer than expected because testing with rack sessions and Roda session_middleware sessions was broken.  You can't use `@app.new(env).csrf_token` in those cases, since there won't be a session handler in `env['rack.session']` (the Roda sessions plugin overrides the session method and doesn't check `env`).

Actually fixing this gets quite involved, due to the multiple types of session and CSRF backends.  I ended up creating a separate Roda app per call to json_request with a route that returns the CSRF token.  I'm open to a more efficient implementation that doesn't depend on implementation details of the CSRF or session backend.

One other change I made is that when the jwt feature is in use, `use_json?` is true if a JWT is provided, even if `use_json?` would not otherwise be true.  I think this is more backwards compatible.

janko:
@jeremyevans Thank you, I really appreciate that you fixed it up 🙏 (and sorry for the extra work 🙈)

The changes makes sense to me. I was wondering whether in the original version fetching the CSRF token would work with sessions plugin if we called `#session` on the Roda object before calling `#csrf_token`. From what I've been looking, the sessions plugin lazily sets `env["rack.session"]`. But your version seems more robust.

PR #146, opened 2021-01-24
Add support for use argon2 for password hashing

AlexeyMatskevich:
* Add the ability to migrate from bcrypt to argon2 and vice versa.
* Add ability to change hash cost when used argon2

Also I think there will be no problem, for example, when some users are still using bcrypt and others are already using argon2 after last migration, and we apply new migration with change the cost.

At the moment, I do not know how best to cover changes with tests for cases of migration from argon2 to bcrypt and when changing the cost for argon2, as well as when checking the password through functions in the database.


lib/rodauth/features/base.rb
```diff
@@ -470,7 +472,22 @@ def database_function_password_match?(name, hash_id, password, salt)
     end
 
     def password_hash_match?(hash, password)
-      BCrypt::Password.new(hash) == password
+      case get_hash_algorithm(hash)
+      when 'bcrypt'
+        BCrypt::Password.new(hash) == password
+      when 'argon2'
+        Argon2::Password.verify_password(password, hash)
+      end
```
AlexeyMatskevich, line 480: It might be worth catching an error here when the hash is invalid and thinking about a more general error. For example, when the user disabled bcrypt and accounts with the old hash remained in the database and ended up in the branch with argon2.

jeremyevans:
I'm not opposed to adding optional support for argon2, but it will need to be more complete than this.  For one, it will need to work with Rodauth's support for storing hashes in a database table that the application does not have select access too.  This means you'll need to modify the database functions in `lib/rodauth/migrations.rb` and make sure they work for argon2 and bcrypt hashes on PostgreSQL, MySQL, and MSSQL.  My understanding is that argon2 is more configurable than bcrypt, which will probably increase the complexity of the functions (on top of the added complexity of supporting both algorithms).  Is that something you will be able to work on?

In terms of implementation, argon2 support should be implemented as a separate feature, instead of being part of the core.

I think the failures in the CI are because `.ci.gemfile` was not modified.

Note that argon2 doesn't offer significant advantages over bcrypt if the attacker does not have direct access to the password hashes.

jeremyevans:
Since this appears to be in a WIP state, I'm going to close this.  We can reopen later when you think it is ready for merging.

AlexeyMatskevich:
At the moment I am almost done with writing the code, I only have a problem with running mssql
```
 % rake db_setup_mssql
sqlcmd -E -e -b -r1 -i spec\sql\mssql_setup.sql
Sqlcmd: 'specsqlmssql_setup.sql': Unexpected argument. Enter '-?' for help.
rake aborted!
Command failed with status (1): [sqlcmd -E -e -b -r1 -i spec\sql\mssql_setu...]
/home/alex/Projects/Opensource/rodauth/Rakefile:137:in `block in <top (required)>'
Tasks: TOP => db_setup_mssql
(See full trace by running task with --trace)
```

PR #147, opened 2021-01-27
Add argon2

AlexeyMatskevich:
I checked that the test scenarios work with Postgres and MySQL, but could not verify for MsSQL. I can't install the tiny_tds gem. Besides, I am sure that I have the wrong regular expression for  MsSQL. Do you have an opportunity to help with MsSQL?


lib/rodauth/migrations.rb
```diff
@@ -131,4 +131,74 @@ def self.create_database_previous_password_check_functions(db, opts={})
...
+SECURITY DEFINER
+SET search_path = #{search_path};
+END
+    when :mysql
+      db.run <<END
+CREATE FUNCTION #{get_salt_name}(acct_id int8) RETURNS varchar(255)
+SQL SECURITY DEFINER
+READS SQL DATA
+BEGIN
+RETURN (SELECT REGEXP_SUBSTR(password_hash, '.argon2id.v=[0-9]+.m=[0-9]+,t=[0-9]+,p=[0-9]+.[^$]+')
```
AlexeyMatskevich, line 167: An alternative to implement the current functionality in the database is to simply select the first 70 characters including the salt, i.e. similar to how it works in rodauth now.

jeremyevans:
Thanks for continuing to work on this.

The approach of using two separate functions to get the database salt is fine.  My original idea was a single function that handled both, so you didn't need multiple round trips, but that isn't a significant concern.

It looks like the MySQL code doesn't work because it depends on a function only available in MySQL 8+.  Can you come up with an approach that will work on MySQL 5+ (which should also work on MariaDB)?  I don't think a selecting a fixed number of characters works for argon2 due to the variable length of parameters.

For MSSQL, the approach mentioned in the first answer to https://stackoverflow.com/questions/39002025/is-there-a-lastindexof-in-sql-server should work, with an additional check that the the password starts with $argon2id$v=.

The sqlcmd failure reported in #146 may be due to the fact that it expects you are running on Windows and uses `\` as the directory separator.  You could try switching `spec\sql\mssql_setup.sql` to `spec/sql/mssql_setup.sql` in the Rakefile and see if that works for you.

If you can't get the tiny_tds gem installed, try testing on JRuby with the jdbc-jtds gem.

I think it would be better to name the feature argon2 instead of argon.

AlexeyMatskevich:
I tried to do according to your advice and install jruby jdbc-jtds and also did raketask to run tests, but now I have problems with roles
```
 % rake db_setup_mssql
rake aborted!
Sequel::DatabaseError: Java::JavaSql::SQLException: CREATE TABLE permission denied in database 'rodauth_test'.
```

jeremyevans:
Can you show how you modified the `db_setup_mssql` rake task to use `jdbc` instead of `tinytds`?  My guess is the JDBC connection string is wrong, and it is trying to use `rodauth_test` instead of `rodauth_test_password` as the user when connecting.  The jdbc adapter doesn't use separate :user/:password options, only a connection string which is passed directly to JDBC.

I probably should modify the tasks so they work with JDBC on JRuby and the existing adapters on CRuby, I'll see if I can work on that later today.

jeremyevans:
I pushed changes in fe5de7950862f6dcdd1a7ce3dfeb131aca43bb4f so that the Rake tasks should work both in CRuby and JRuby.

AlexeyMatskevich:
Thanks for help. Answering your question, I used this code to connect in `db_setup_mssql` rake task:
```
  Sequel.connect('jdbc:jtds:sqlserver://0.0.0.0:1433/rodauth_test;user=rodauth_test;password=Rodauth1.') do |db|
    Sequel::Migrator.run(db, 'spec/migrate')
  end
  Sequel.connect('jdbc:jtds:sqlserver://0.0.0.0:1433/rodauth_test;user=rodauth_test;password=Rodauth1.') do |db|
    Sequel::Migrator.run(db, 'spec/migrate_password', :table=>'schema_info_password')
  end
```
Now with changes in fe5de79 I was able to run migrations successfully!
I added locally this rake task to run tests
```
desc "Run specs on MSSQL"
task :spec_mssql do
  sep = ';' if RUBY_ENGINE == 'jruby'
  spec.call('RODAUTH_SPEC_DB'=>"#{adapters[:mssql]}://0.0.0.0:1433/rodauth_test#{sep || '?'}user=rodauth_test#{sep || '&'}password=Rodauth1.")
end
```
In fact, I am using MsSQL running in docker, I took the image from the official repository https://hub.docker.com/_/microsoft-mssql-server. For this reason my database address is `0.0.0.0:1433`. In the image database version from 2017 year.
And so I was able to run the tests.
Unfortunately, many tests fail.
I am attaching a gist with their results https://gist.github.com/AlexeyMatskevich/7915197077022469f80d98070c05fc0f

jeremyevans:
Note that the way the migrations are run (with user `rodauth_test` instead of `rodauth_test_password`) removes all the security.  You want the password table owned by a different database user than the user running the application (tests in this case), otherwise the database user can just read the table to get the password hashes if the application is compromised.  However, that should not affect your test results.

Not sure why tests are resulting in so many failures. Rodauth's tests run fine with SQL Server Express (I think 2017) on a Windows 10 machine, as that's how I test locally.  Maybe SQL Server on Linux or maybe docker is changing something.

The tests do show issues with the argon2 support on MSSQL (invalid salt errors), so you should probably focus on addressing those issues.

I would recommend using the demo site for testing locally.  Test with argon2 support enabled by default, make sure you can create accounts and login with them and that the password hash table is storing the argon2id hashes.  When that works correctly for you, please let me know and I can work on testing/merging this PR.

AlexeyMatskevich:
Yes, the presence of unsuccessful cases in the tests does not prevent me from concentrating on working on my feature with argon2, I will try to add the correct functions to the database and test local.

AlexeyMatskevich:
Hi, I used a windows 10 computer to work and am developing on ubuntu in the WindowsOnLinux subsystem. So I tried to install mssql on the host machine (windows) and use it from the wsl environment. I managed to do it successfully, after that all tests passed successfully both on mri and jruby. 
I changed the function code to a working state and abandoned the algorithm with two requests to db for salt.

jeremyevans:
Sounds good, thanks for getting all of the tests working.  I'll starting testing this on Monday.

jeremyevans:
I was able to start testing this today. Tests pass on PostgreSQL, MSSQL, and SQLite.  Tests fail locally for me on MySQL: `Sequel::DatabaseError: Mysql2::Error: Regex error 'POSIX collating elements are not supported at offset 1'`.  This appears to be due to the use of MariaDB (https://stackoverflow.com/questions/45595058/mysql-regex-with-character-doesnt-work-anymore-posix-collating-elements-a).  Can you modify the MySQL function so it works on both MySQL 5+ and MariaDB?  Maybe a similar approach to that used by MSSQL may work.

One issue I noticed is this doesn't currently work with the disallow_password_reuse feature.  Maybe we should refactor this support so that the default `rodauth_get_salt` function supports argon2 if passed an `argon2` option?  That way a separate function for argon2 support is not needed.  I think that would fix the issue and probably be simpler overall.

AlexeyMatskevich:
If I understood the idea correctly, then I implemented it

jeremyevans:
Sorry that I didn't explain what I wanted well.  I can see now that my previous comment was easy to misinterpret.  What I meant was, change  `create_database_authentication_functions` to accept an `:argon2` option, and if so, have the `rodauth_get_salt` function defined check for both bcrypt and argon2id hashes (if the `:argon2` option is not passed, just use the current function code that only supports bcrypt).

Also, please do not make unnecessary style changes when submitting code changes, as it makes the code significantly more difficult to review.

jeremyevans:
I did test the new MySQL function for getting the salt and it is working for me, so that is great!

AlexeyMatskevich:
Now the idea is clear to me. I will be able to make changes including stylistic changes and finish the migration code today.

jeremyevans:
Awesome!  Thank you for your continued work on this, and I apologize for not being clear earlier.

jeremyevans:
Just to keep you updated, I've been working on this today.  I'm doing some refactoring and adding documentation, but everything so far is working as expected.  I want to add the ability to run the entire test suite using the argon2 feature just to make sure that all parts of Rodauth work with it.  Once I've added that, if it doesn't show any issues, I'll be committing this support.

PR #150, opened 2021-02-22
introduce `#token_param` to compute the verification token (prefixed …

apotonick:
introduce `#token_param`  to compute the verification token (prefixed with the user id).
this was done directly in the email link rendering before.

Could you help me where to add a test for this?

### DIFF
--- a/lib/rodauth/features/email_base.rb
+++ b/lib/rodauth/features/email_base.rb
@@ -51,7 +51,11 @@ module Rodauth
     end
 
     def token_link(route, param, key)
-      route_url(route, param => "#{account_id}#{token_separator}#{convert_email_token_key(key)}")
+      route_url(route, param => token_param(key))
+    end
+
+    def token_param(key)
+      "#{account_id}#{token_separator}#{convert_email_token_key(key)}"
     end
 
     def convert_email_token_key(key)
@@ -71,7 +75,6 @@ module Rodauth
           return
         end
       end
-
       ds = account_ds(id)
       ds = ds.where(account_status_column=>status_id) if status_id && !skip_status_checks?
       ds.first


jeremyevans:
Overriding the new `token_param` method and changing the token format would break parsing of tokens by default.  So I'm not sure we want to make this easier to override.  What is the use case?

apotonick:
Here's an example use case: https://github.com/apotonick/rodauth/blob/b14a4e3fa3a1590641dbabfdd66db93167cbdadd/test/api_test.rb#L230

```ruby
  def setup_account_verification
    super()
    return account, @verify_account_key_value, token_param(@verify_account_key_value)
  end
```
In Tyrant, we want to compile the email for "reset password" manually, so the Rodauth link rendering is a bit in the way.

Please excuse the messy "test", I'm still prototyping but so far I can use Rodauth without Roda and Sequel for 
* creating account with scrambled password
* verifying account using token
* resetting password
* updating password using token

:v: pretty exciting as it saves us a lot of work in Tyrant!

jeremyevans:
I'm OK with extracting the method.  `token_param` sounds like it would return the name of a parameter as opposed to the value of a parameter, so I've renamed it to `token_param_value`.  Cherry-picked at 82e4801879c57abacd17efd13c2fafe01559ccc4 , rename at bdc57b3803dc08b738f7d76d862f42d8424fcf2b .

You should be aware that the method is still private and considered an implementation detail.  That being said, it's not likely to change.

apotonick:
Oh, thanks Jeremy! :heart: 

I have a million of questions (and - maybe stupid - ideas). I will commit a document in my repo so you and @janko can use Github's commenting. Looking forward to your opinions! :beers:

PR #151, opened 2021-02-23
Improve Rodauth::Auth inheritance

janko:
I'm working on adding support to rodauth-rails for defining `Rodauth::Auth` subclasses directly, and hooking them into the Roda app. Essentially something like this:

```rb
  configure do
    # ... main user logic ...
  end
end
```
```rb
class RodauthAdmin < Rodauth::Auth
  configure do
    # ... admin logic ...
  end
end
```
```rb
class RodauthApp < Roda
  opts[:rodauths] = {
    nil    => RodauthMain,
    :admin => RodauthAdmin,
  }

  plugin :rodauth do
    # nothing, just to load the plugin
  end
end
```

This kind of setup would provide several advantages for the user:

* authenication logic for different configurations can be easily separated into different files (and separated from routing logic)

* common logic can be shared between configurations using inheritance

  ```rb
  class RodauthBase < Rodauth::Auth
    # common logic
  end
  class RodauthMain < RodauthBase
    # ...
  end
  class RodauthAdmin < RodauthBase
    # ...
  end
  ```

* it's more natural to define additional methods on the Rodauth object

  ```rb
    configure do
      # this works, but it's a bit strange and nested
      auth_class_eval do
        def admin?
          # ...
        end
      end
    end

    # this is regular way how methods are defined, which is familar to everyone
    def admin?
      # ...
    end
  end
  ```

* the `Rodauth::Auth` subclass is named with a constant instead of being an anonymous class, which makes for better exception reports in `NoMethodError` situations

To make multiple levels of inheritance work, I've modified `Rodauth::Auth` to initialize class instance variables upfront, and assign a copy of them to the subclass. With that I also wanted to make loading the same feature multiple times work, in case the user wants to tweak an already loaded feature in the subclass (I also almost needed this behaviour in rodauth-rails regardless of the inheritance-related changes). The commit messages provide more details.

### DIFF
--- a/lib/rodauth.rb
+++ b/lib/rodauth.rb
@@ -295,9 +295,8 @@ module Rodauth
 
     def enable(*features)
       features.each do |feature|
-        next if @auth.features.include?(feature)
         load_feature(feature)
-        @auth.features << feature
+        @auth.features << feature unless @auth.features.include?(feature)
       end
     end
 
@@ -309,7 +308,7 @@ module Rodauth
       enable(*feature.dependencies)
       extend feature.configuration
 
-      @auth.routes.concat(feature.routes)
+      @auth.routes |= feature.routes
       @auth.send(:include, feature)
     end
   end
--- a/lib/rodauth.rb
+++ b/lib/rodauth.rb
@@ -257,18 +257,21 @@ module Rodauth
   class Auth
     class << self
       attr_accessor :roda_class
-      attr_reader :features
-      attr_reader :routes
+      attr_accessor :features
+      attr_accessor :routes
       attr_accessor :route_hash
     end
 
+    @features = []
+    @routes = []
+    @route_hash = {}
+
     def self.inherited(subclass)
       super
-      subclass.instance_exec do
-        @features = []
-        @routes = []
-        @route_hash = {}
-      end
+      subclass.roda_class = roda_class
+      subclass.features = features.dup
+      subclass.routes = routes.dup
+      subclass.route_hash = route_hash.dup
     end
 
     def self.configure(&block)
--- a/spec/rodauth_spec.rb
+++ b/spec/rodauth_spec.rb
@@ -1,6 +1,25 @@
 require_relative 'spec_helper'
 
 describe 'Rodauth' do
+  it "should support configuring a feature multiple times" do
+    require "rodauth"
+
+    auth_class = Class.new(Rodauth::Auth)
+    auth_class.roda_class = Class.new(Roda)
+    auth_class.configure do
+      enable :login
+      login_redirect "/"
+    end
+
+    auth_class.configure do
+      enable :login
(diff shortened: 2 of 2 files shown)


jeremyevans:
How about just an `:auth_class` plugin option to allow users to specify their own class (used directly, instead of subclassing `Auth`)? I think that would avoid needing to have all of this extra complexity in Rodauth itself.  If an external library such as rodauth-rails wants to enable feature inheritance, they can just have an `Auth` subclass that adds that feature, and use that as a superclass of `:auth_class` classes.

janko:
Yes, an `:auth_class` option would definitely make sense, it would address modifying `opts[:rodauths]` directly as in my example, so it would make it easier to use a custom `Auth` subclass.

The reason I wanted to uptream the inheritance changes is so that any Rodauth user can benefit from them, as it satisfies a use case that's not specific to rodauth-rails. But I'm totally fine moving these changes to rodauth-rails; I already have a `Rodauth::Rails::Auth` class that subclasses `Rodauth::Rails`, so I can add the inheritance logic there.

The only thing I would need, if I wanted to allow users to configure the same feature in both a superclass and a class (e.g. when having `RodauthMain < RodauthBase < Rodauth::Rails::Auth`), is the `Configuration` change to call `load_feature` even if the feature is already loaded, as I cannot easily modify that in rodauth-rails. Would you consider accepting this change into Rodauth? I've already made it as a separate commit that should be cherry-pickable.

I could then make a separate PR that adds the `:auth_class` plugin option.

jeremyevans:
I think a better approach to handle configuring the plugin multiple times would be to cache the Configuration instance that Auth uses (creating it in `inherited`) and reuse it, instead of creating a new one in every call to `configure`.   What are your thoughts on that approach?

janko:
This actually came to mind, and I liked that this way the features that were already loaded would retain their configuration DSL if `configure` happens to be called again. Good to hear that you would prefer it, I can open a PR for that.

One thing I'm not clear with, when I will want to add `Rodauth::Auth` inheritance improvements to rodauth-rails, what should I do with the `Configuration` instance so that inheritance is respected? Subclassing the superclass' `Configuration` class wouldn't work, because the configuration methods are defined only on a `Configuration` instance's singleton class. I would be fine with just creating a new `Configuration` instance, but if I'm planning to modify `inherited` so that it *copies* the `@auth.features` array instead of resetting it (for the reasons described in this PR), then if I have configured a feature in `A < Rodauth::Auth`, configuring it in `B < A` again wouldn't work, because the `Configuration` object would skip loading the feature given that it's already present in `@auth.features`, so it wouldn't get extended with that feature's configuration methods.

Sorry, this turned out to be a more involved than I thought, this should probably have been discussed on the google group first 😛

jeremyevans:
You can clone the Configuration instance in `inherited`, that should allow things to continue to work.

PR #153, opened 2021-02-24
Add :auth_class plugin option for providing a specific Rodauth::Auth subclass

janko:
As discussed in https://github.com/jeremyevans/rodauth/pull/151, this PR adds an `:auth_class` plugin option that allows providing a custom `Rodauth::Auth` subclass. Let's say the user has the `Rodauth::Auth` subclasses defined as follows:

```rb
  configure do
    # ...
  end
end
```
```rb
class RodauthAdmin < Rodauth::Auth
  configure do
    # ...
  end
end
```

Previously it was possible to specify these classes by overriding `opts[:rodauths]`:

```rb
class RodauthApp < Roda
  opts[:rodauths] = {
    nil => RodauthMain,
    :admin => RodauthAdmin,
  }

  plugin :rodauth do
    # just to load the plugin
  end
end
```

The new `:auth_class` option makes this easier:

```rb
class RodauthApp < Roda
  plugin :rodauth, auth_class: RodauthMain do
    # ..
  end

  plugin :rodauth, name: :admin, auth_class: RodauthAdmin do
    # ...
  end
end
```

### DIFF
--- a/README.rdoc
+++ b/README.rdoc
@@ -852,6 +852,9 @@ which configures which dependent plugins should be loaded. Options:
          still need to load the render plugin manually.
 :name :: Provide a name for the given Rodauth configuration, used to
          support multiple Rodauth configurations in a given Roda application.
+:auth_class :: Provide a specific `Rodauth::Auth` subclass that should be set
+               on the Roda application. By default, an anonymous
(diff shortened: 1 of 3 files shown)


jeremyevans:
Looks good, thanks!

PR #155, opened 2021-03-02
Use new-password autocomplete value for password field when resetting password

basabin54:
Set @password_field_autocomplete_value = 'new-password' for the reset-password feature


jeremyevans:

basabin54:
Happy to help! 👍

PR #157, opened 2021-03-08
Allow not passing a block when loading the plugin

janko:
When using the approach of passing `Rodauth::Auth` subclasses explicitly via the `:auth_class` option, the auth class would be configured already, so in this case it would be convenient to be able to skip passing the block when loading the plugin.

```rb
  configure do
    # ...
  end
end
```
```rb
plugin :rodauth, auth_class: RodauthMain
```

I was thinking whether I should move this into `Rodauth::Auth.configure`, or still require the block when `:auth_class` is not passed in, so that it's not possible to set up a `Rodauth::Auth` subclass without the `:base` feature loaded. What are your thoughts on this?

### DIFF
--- a/lib/rodauth.rb
+++ b/lib/rodauth.rb
@@ -46,7 +46,7 @@ module Rodauth
       auth_class = app.opts[:rodauths][opts[:name]] = Class.new(auth_class)
       auth_class.roda_class = app
     end
-    auth_class.configure(&block)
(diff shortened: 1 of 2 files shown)


jeremyevans:
I think the current implementation is fine.  If someone loads the plugin without a block, it shouldn't cause any harm.   I'll test and merge later today.

PR #159, opened 2021-04-10
Allow setting additional values to the active session record

janko:
This allows overriding the active session insert hash to add additional values, which makes it easier to assign things such as IP address, device type, or country/city to active sessions, for giving the user a better overview of their active sessions.

```rb
active_session_insert_hash do
  super.merge(
    ip: request.ip,
    user_agent: request.user_agent,
    ...
  )
end
```

### DIFF
--- a/doc/active_sessions.rdoc
+++ b/doc/active_sessions.rdoc
@@ -37,9 +37,13 @@ inactive_session_error_status :: The error status to use when a JSON request is 
 session_id_session_key :: The session key name to use for storing the session id.
 session_inactivity_deadline :: The number of seconds since last use after which the session will be considered expired (1 day by default).  Can be set to nil to not check session inactivity.
 session_lifetime_deadline :: The number of seconds since session creation after which the session will be considered expired (30 days by default).   Can be set to nil to not check session lifetimes.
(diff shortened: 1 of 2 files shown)


jeremyevans:
Thanks for the patch! This looks good to me.  I'll test and merge on Monday.

janko:
As discussed on the google group, I've also added an `active_sessions_update_hash` method for updating the currently active session with current data. This is useful for keeping the extra session data up-to-date on each request, e.g. updating the IP address when it has changed.

I wanted it to also be possible to do these updates if session inactivity deadlines are disabled. As I see it, even without the new configuration methods for adding extra data, it might still be useful to keep the `last_use` column up-to-date. So I added a `update_current_session?` configuration method which defaults to true when `session_inactivity_deadline` is set.

Because I sensed that we might want to keep the `active_sessions_*` naming convention, I've renamed the original `active_session_key_insert_hash` and `active_session_key_value` methods to `active_sessions_insert_hash` and `active_sessions_key`.

janko:
@jeremyevans Thanks a lot for such a quick review and merge 🙏🏻

PR #160, opened 2021-04-16
Fix typo in unlock account email template

janko:


jeremyevans:
Thanks for the patch!

PR #162, opened 2021-04-22
Set a reason code when an error happens

renchap:
For example, this allows you to add this code to the JSON response so your app can display a custom screen or message when certain kinds of errors happen.

I updated every `throw_error_status` and `set_response_error_status` to set the reason code, with special handling for the login and password requirements.

The error reason code is added to the JSON response in `spec_helper.rb` so we can match it in tests.
Tests are passing locally using PostgreSQL.


jeremyevans:
Thank you very much for working on this.  I'll try to review it later today.

jeremyevans:
Squash merged at 2544d9653b92953812db857b1f206c84ca4462c0.  Thanks!

PR #163, opened 2021-05-02
More set_error_reason support

renchap:
This is a follow up to #162

`set_error_reason` was not called on redirections It is now set everywhere `set_redirect_error_status` is called.

Documentation has been updated with the new reasons, I tried to reuse the existing ones where possible.


jeremyevans:
Thanks for working on this! This looks good.  I have a couple minor changes to the error reasons, and a few minor fixes, but as soon as testing finishes I'll merge that.

PR #164, opened 2021-05-06
Update templates for Boostrap 5

janko:
[Bootstrap 5.0 was released](https://blog.getbootstrap.com/2021/05/05/bootstrap-5/) yesterday. This PR updates the templates with some of the changes:

* label tags now require the `form-label` class
* `form-group` class has been removed, the official docs use `mb-3` instead, so we use them here as well


jeremyevans:
Thanks, this looks good and should be backwards compatible as it doesn't remove any existing classes. Will merge shortly.

PR #165, opened 2021-05-07
Make jwt_refresh work with active_sessions when allowing expired JWTs for refresh

renchap:
If `rodauth.check_active_session` is called before `r.rodauth` then `@jwt_refresh_route` is not yet set and the JWT is decoded with `:verify_expiration: true`, which fails even if we want to allow expired access tokens to be provided for refresh.

As discussed on IRC this seems to be the best way to detect if the request is for the refresh route while not beeing in the route handler itself.

The added test is failing without this patch, and green with it.

### DIFF
--- a/lib/rodauth/features/jwt_refresh.rb
+++ b/lib/rodauth/features/jwt_refresh.rb
@@ -32,6 +32,8 @@ module Rodauth
     )
 
     route do |r|
+      # For backward compatibility, unused in core Rodauth
+      # RODAUTH3: Remove
       @jwt_refresh_route = true
       before_jwt_refresh_route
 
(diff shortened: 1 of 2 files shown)


lib/rodauth/features/jwt_refresh.rb
```diff
@@ -32,7 +32,6 @@ module Rodauth
     )
 
     route do |r|
-      @jwt_refresh_route = true
```
jeremyevans, line 35: Please keep this for backwards compatibility, maybe with a `# RODAUTH3` comment added to mark it as something we can drop in the next major version.

jeremyevans:
This looks good.  I'll try merging on Monday.

jeremyevans:
Cherry-picked at 1843f9b70162e938dedc2bdbcdfe8fc1ca5962bb

PR #166, opened 2021-05-21
Add `remembered_session_id` to check if remembered

bjeanes:
Hey @jeremyevans,

I have a use-case where I need to check if the session has a valid remember cookie set but currently the code which validates the cookies presence and contents is buried inside `load_memory` which both has side effects and early exits in the presence of a valid logged in session.

I'm hoping you won't be opposed to an extraction like the one herein. Let me know if you'd prefer a different name or it broken up a different way.


jeremyevans:
I'm OK adding a method for this.  A few issues with the current commit:

* Do not include unrelated whitespace changes.
* Do not modify CHANGELOG.
* Add an entry for the new configuration method in the appropriate file in the `doc` folder.

bjeanes:
Whoops sorry about those whitespace changes, in particular—I had excluded them in my first commit but then amended it and accidentally included.

I've pushed a new commit which adds the docs, leaves the `CHANGELOG` untouched, and preserves existing whitespace.

jeremyevans:
This looks good.  I've already done full prerelease testing for tomorrow's release, so I'll test and merge this on Monday.

jeremyevans:
Cherry-picked at 3ea500a1412e9fb9e41c61162616d1f8af62d5b5

PR #169, opened 2021-07-02
Remove registration of internal rodauth class

janko:
As discussed on the google group, this modifies instantiating of internal rodauth instance in a way that doesn't require the internal rodauth class to be registered on the roda class. This helps it remain invisible when introspecting the roda class.


jeremyevans:
This looks good, thanks!

PR #170, opened 2021-07-02
Add :params option for custom parameters to internal_request feature

janko:
This allows the caller to pass additional custom parameters that are not known to Rodauth to internal requests. For example, if we've added an additional `name` field to the create account form for the user to enter their full name, we might want to pass the `name` parameter into the internal request as well. We can now do this via `:params`:

```rb
App.rodauth.create_account(
  login: "user@example.com",
  password: "secret",
  params: { "name" => "John Smith" }
)
```


jeremyevans:
This looks good.  Thanks for working on it!

jeremyevans:
Cherry-picked at 7f2a37780210694160fa418cf006109b99849739

PR #171, opened 2021-07-04
Use configured domain in email links and other URLs

janko:
Rodauth allows configuring the domain via the `#domain` configuration method:

```rb
domain "example.com"
```

However, this domain wasn't used when generating full URLs via `#route_url`, because that uses `Rack::Request#base_url`, which reads the host from the rack env. This becomes an issue when using the internal_request feature, which sets the host in rack env to an invalid domain that is expected to be overridden. Setting the `#domain` gets rid of the error, but email links will still use the invalid placeholder domain.

This PR fixes the issue by modifying `#base_url` to use the configured domain. To retain backwards compatibility with proxies, we also change `#domain` to read from `Rack::Request#authority`, which is what `Rack::Request#base_url` reads from as well.

Let me know if you agree with this approach. I've also considered modifying the `env` hash with `#domain` value after instantiating the Rodauth instance in the `.internal_request` method. However, I felt that email links and other URLs should use the configured `#domain` even when not using the internal_request feature.

### DIFF
--- a/lib/rodauth/features/base.rb
+++ b/lib/rodauth/features/base.rb
@@ -448,7 +448,9 @@ module Rodauth
     end
 
     def base_url
-      request.base_url
+      url = "#{request.scheme}://#{domain}"
+      url << ":#{request.port}" if request.port != Rack::Request::DEFAULT_PORTS[request.scheme]
+      url
     end
 
     def domain
--- a/lib/rodauth/features/internal_request.rb
+++ b/lib/rodauth/features/internal_request.rb
@@ -234,6 +234,7 @@ module Rodauth
          "SCRIPT_NAME" => "",
          "HTTP_HOST" => INVALID_DOMAIN,
          "SERVER_NAME" => INVALID_DOMAIN,
+         "SERVER_PORT" => 443,
          "CONTENT_TYPE" => "application/x-www-form-urlencoded",
          "rack.input"=>StringIO.new(''),
          "rack.url_scheme"=>"https"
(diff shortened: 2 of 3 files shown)


janko:
Ok, I see this will be bit more involved on Rack < 2. I will first wait for a confirmation of this approach before addressing the Rack < 2 compatibility.

jeremyevans:
I'm OK switching to this approach, so you only need to modify `domain` instead of having to modify both `domain` and `base_url`.  So I guess we just need to figure out how to get it working on Rack 1.

janko:
I've changed the initial implementation of `#domain` to continue reading from `Rack::Request#host`, because that takes proxies into account as well. `Rack::Request#authority` includes a port number if it's a non-default one, so that could have broken backwards compatibility.

The current implementation is the same as `Rack::Request#base_url` in Rack 1.x (with `#domain` instead of `Rack::Request#host`), and it also turns out to be the most suitable for Rack 2.x. For Rack 1.x I've needed to set `SERVER_HOST`, otherwise `Rack::Request#port` returns `0` because [it doesn't handle missing `SERVER_HOST`](https://github.com/rack/rack/blob/47a1fd73fc77f094573f4215b0fc884f4ac1c03c/lib/rack/request.rb#L102). I've verified that Puma indeed sets this, and I believe other web servers set it too.

jeremyevans:
This looks good.  I'll test and merge tomorrow.

`SERVER_PORT` is supposed to be optional according to the rack SPEC.  I'm guessing in Rack 1.x it was either non-optional or there was a bug expecting it to be present.

PR #172, opened 2021-07-10
Expose flash, reason, and field errors on InternalRequestError

janko:
To give the user all available information about why an internal request failed, we include flash, reason, and field errors in the exception message, and also expose this data via attributes on the exception object.


jeremyevans:
Thanks for working on this. This looks good.  I'll merge and test on Monday.

PR #174, opened 2021-07-13
Implement `#lock_account` without setting `@account`

janko:
This is a really small thing, but I noticed that `#lock_account` in internal_request feature can be implemented without setting `@account` (at least without any test failures), so I thought it's worth the change.


jeremyevans:
Thanks for the patch!  This definitely looks simpler.

PR #176, opened 2021-07-28
Make internal requests work in JSON-only mode

janko:
When `#only_json?` is set to true, the JSON feature requires that requests have `Content-Type` and `Accept` headers set to JSON. This is not the case in internal requests, which have a form-data content type set up.

Since in internal requests we don't want to deal with HTTP-level requirements, we set `#only_json?` to false in the internal auth class to allow non-JSON requests. This should be fine, because in internal requests params are grabbed from an internal hash anyway.

### DIFF
--- a/lib/rodauth/features/internal_request.rb
+++ b/lib/rodauth/features/internal_request.rb
@@ -92,6 +92,10 @@ module Rodauth
       @internal_request_return_value
     end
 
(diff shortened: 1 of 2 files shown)


jeremyevans:
I agree, it makes sense to do this.  Thanks for the patch!

PR #178, opened 2021-09-22
Return error status when unverified account exists on account creation

janko:
In #177 I missed another place in verify_account where error status should be set, which is when an unverified account exists when attempting to create an account.

I've switched to using `#set_response_error_reason_status`, as that's used in most other places. I've also added a test for the lockout feature that asserts that a 4xx response status is set when showing the lockout page.


jeremyevans:
Thanks for the patch.  This also looks good.  I agree that setting the error reason is the way to go in this case.

PR #181, opened 2021-11-08
Set configuration name for provided auth classes

janko:
This ensures that auth classes that have been defined explicitly will also have configuration name assigned, just like anonymous classes.

I came across this when building an i18n integration, which uses the configuration name for translation namespaces.

### DIFF
--- a/lib/rodauth.rb
+++ b/lib/rodauth.rb
@@ -50,13 +50,14 @@ module Rodauth
     else
       json_opt != :only
     end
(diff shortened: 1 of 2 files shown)


jeremyevans:
Thanks for the patch! I don't think this will work correctly if you use the same auth class in multiple configurations with different names.  However, I don't think anything will handle that case correctly, and this does a better job of handling the common case, so I agree we should go this direction.

jeremyevans:
Cherry-picked as 0fc8280f4515ee3769c610fdf0bbb1e1ff1f5600

PR #187, opened 2021-12-09
Assign internal request subclass to a constant

janko:
For named auth classes, internal request subclass is still anonymous:

```rb
end

class RodauthApp < Roda
  plugin :rodauth, auth_class: RodauthMain do
    enable :internal_request
  end
end

RodauthMain.internal_request_eval do
  self.class #=> #<Class:0x0000000143b57680>
end
```

I thought it would be nice for introspection and error messages if the internal subclass was assigned into a constant for named auth classes. With the changes in this pull request, we now have:

```rb
RodauthMain.internal_request_eval do
  self.class #=> RodauthMain::InternalRequest
end
```

This is especially useful with rodauth-rails, which exposes an `instance` class method that returns an internal request instance:

```rb
rodauth = RodauthMain.instance
rodauth #=> #<RodauthMain::InternalRequest ...>
```

We make the constant private to ensure it stays internal.

### DIFF
--- a/lib/rodauth/features/internal_request.rb
+++ b/lib/rodauth/features/internal_request.rb
@@ -366,6 +366,11 @@ module Rodauth
           end
         end
       end
+
+      if klass.name
+        klass.const_set(:InternalRequest, internal_class)
+        klass.private_constant :InternalRequest
+      end
     end
   end
 end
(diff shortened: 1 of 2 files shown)


jeremyevans:
This seems fine.  I've merged it locally.  I'm going to always set the constant, since there is no problem setting the constant even if the auth class is anonymous.

PR #190, opened 2021-12-13
Ensure roda class is assigned before calling `#post_configure`

janko:
It's possible to define named auth classes separately from plugin initialization, and configure the features directly via `Rodauth::Auth.configure`. Most features work just fine without roda class being assigned, but internal_request feature needs that when creating the internal auth subclass in `#post_configure`. Third-party plugins might rely on this too, some may even try to load Roda plugins.

To ensure roda class is always assigned, we call `#post_configure` only in the roda plugin's configure method. That way it will be called only after the auth class is has been registered on the Roda app.

```rb
  configure do
    enable :internal_request
  end
end

class RodauthApp < Roda
  plugin :rodauth, auth_class: RodauthMain
end

RodauthMain.account_exists?(login: "user@example.com")
# BEFORE: fails when trying to initialize the roda class
# AFTER: works without any issues
```

This is a follow-up to [this discussion](https://github.com/jeremyevans/rodauth/discussions/186).


jeremyevans:
This looks good on initial review.  I'll try to merge and test this tomorrow.

PR #191, opened 2021-12-19
Support multi-level inheritance

janko:
I've initially suggested this in #151, where we instead ended up making minimal changes that allowed rodauth-rails to support multi-level inheritance. With the support of named auth classes (which rodauth-rails 1.0 will now generate by default), I wanted to make another attempt at supporting multi-level inheritance in Rodauth itself. I think it's a pretty convenient way to share common configuration, and I thought it would be nice for Rodauth to provide this convenience out-of-the-box.

The internal_request feature already implements this functionality, so here we're just moving that into the core. Unlike my original attempt, here I didn't change any attribute readers/writers/accessors, nor assigned instance variables directly on the `Rodauth::Auth` class.

### DIFF
--- a/doc/guides/share_configuration.rdoc
+++ b/doc/guides/share_configuration.rdoc
@@ -1,0 +1,30 @@
+= Share configuration via inheritance
+
+If you have multiple configurations that needs to share some amount of
+authentication behaviour, you can do so through inheritance. For example:
+
+  require "rodauth"
+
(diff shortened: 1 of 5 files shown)


jeremyevans:
Thanks for the patch!  I'm OK with adding this feature.  Let me play around with this, I might be able to make changes to simplify the implementation slightly.

janko:
Nice, thank you 🙏🏻

PR #195, opened 2021-12-20
Pass block directly to internal_request_eval handler

janko:
I was looking at the inspect output of the internal request instance for a named auth class when using `internal_request_eval`, and I noticed it was busier than I expected:

```
#<RodauthMain::InternalRequest:0x0000000138feb710
 @internal_request_block=
  #<Proc:0x0000000138ff09e0 /Users/janko/.rbenv/versions/3.0.1/lib/ruby/gems/3.0.0/bundler/gems/rodauth-rails-9152ba36c4c5/lib/rodauth/rails.rb:29>,
 @internal_request_return_value=#<RodauthMain::InternalRequest:0x0000000138feb710 ...>,
 @params={},
 @scope=#<RodauthApp:0x0000000138feaec8 @_request=#<RodauthApp::RodaRequest POST />, @_response=#<RodauthApp::RodaResponse nil {} []>>,
 @session={}>
```

In rodauth-rails, there is a wrapper around `internal_request_eval` that returns the internal request instance, so it would be easy for me to remove `@internal_request_block` and `@internal_request_return_value` instance variables there before returning the instance.

However, it didn't seem to me like it was necessary to assign the block passed to `internal_request_eval` to an attribute accessor, so I made the change to pass the block down directly through the method chain. Let me know what you think of this change.

### DIFF
--- a/lib/rodauth/features/internal_request.rb
+++ b/lib/rodauth/features/internal_request.rb
@@ -36,7 +36,6 @@ module Rodauth
     attr_accessor :session
     attr_accessor :params
     attr_reader :flash
-    attr_accessor :internal_request_block
 
     def domain
       d = super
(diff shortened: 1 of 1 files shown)


jeremyevans:
The block deliberately set to an attribute, so that any time while handling an internal request, you can call the block to get per-request behavior.  I guess there isn't an example of using this in the specs, but there probably should be.

janko:
I see, that makes sense, I didn't realize that was a feature for other internal request methods 👍🏻

PR #209, opened 2022-02-04
Fix spelling of global_logout_label default in active_sessions feature

sterlzbd:
Noticed this while adding active_sessions to a project. This seemed minor enough that we don't need to add a spec for it but I can if you think we should!


jeremyevans:
Thanks for fixing this.  I think we should have active_sessions_spec check the box using the label instead of the id as it does currently, as a way of testing this.

sterlzbd:
Thanks!

PR #211, opened 2022-02-10
Clear session when unverified grace period expired on require login

janko:
*Follow-up to the discussion in https://github.com/jeremyevans/rodauth/discussions/210*

Currently a user whose unverified grace period expired can still continue to browse the application, as long as they don't close their browser or visit any authentication routes.

We address that by adding the grace period expiration to the session, and logging the user out when it has expired. By storing this in session, we avoid a database query on each call to require_login. We maintain backwards compatibility with existing session values still set to `true`.

### DIFF
--- a/lib/rodauth/features/verify_account_grace_period.rb
+++ b/lib/rodauth/features/verify_account_grace_period.rb
@@ -30,10 +30,17 @@ module Rodauth
       false
     end
 
+    def require_login
+      if unverified_grace_period_expired?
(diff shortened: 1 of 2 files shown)


jeremyevans:
Looks good! I'll test and merge shortly.

jeremyevans:
Cherry-picked at 859de3a9e3a74a1449a881f7d269516ff8c25831

PR #212, opened 2022-02-12
Make `#require_account` method public

janko:
*Follow-up to the discussion in https://github.com/jeremyevans/rodauth/discussions/210*

This allows ensuring the account exists in the database before proceeding with the request. This is useful for preventing errors in development when the account record has been manually deleted, or if the account record has been deleted in production for whatever reason, and the performance impact of fetching the account record on each request is acceptable.

### DIFF
--- a/README.rdoc
+++ b/README.rdoc
@@ -990,6 +990,10 @@ require_authentication :: Similar to +require_login+, but also requires
                           two factor authentication.  Redirects the request to
                           the two factor authentication page if logged in but not
                           authenticated via two factors.
+require_account :: Similar to +require_authentication+, but also loads the logged
+                   in account to ensure it exists in the database.  If the account
+                   doesn't exist, or if it exists but isn't verified, the session
+                   is cleared and the request redirected to the login page.
 logged_in? :: Whether the session has been logged in.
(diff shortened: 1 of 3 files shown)


jeremyevans:
Looks good, thanks for working on this.

janko:
@jeremyevans I wanted this method to be a drop-in replacement for `#require_login` and `#require_account`, but I just found one edge case when using `verify_account` without `verify_account_grace_period`. There we expect an unverified account to be allowed to use the app until the session expires, however, `#account_from_session` that gets called will only find verified accounts, and thus would immediately logout the user that just created the account (assuming `#require_account` was called).

jeremyevans:
`verify_account` sets `create_account_autologin?` to `false`, so I assume this is only a problem if a user sets it to `true`.  In that case, the user needs to choose whether they want to require accounts or whether they want to allow autologin for unverified users when accounts are created.

janko:
Thanks for merging so promptly 🙂

> verify_account sets create_account_autologin? to false, so I assume this is only a problem if a user sets it to true

Oh, right, I thought autologin after account creation stays on by default. OK, since I don't expect turning autologin on will be common, I agree it's a decent trade-off 👍🏻

PR #222, opened 2022-03-04
Fix bug with update_hash_ds

bjeanes:
I'm surprised there wasn't a test that caught this for the `sms_codes` feature, which uses it thusly:

https://github.com/jeremyevans/rodauth/blob/d21277ddaae0b74ba70c86a8f08f467344085f76/lib/rodauth/features/sms_codes.rb#L495-L497

### DIFF
--- a/lib/rodauth/features/base.rb
+++ b/lib/rodauth/features/base.rb
@@ -756,7 +756,7 @@ module Rodauth
       num = ds.update(values)
       if num == 1
         values.each do |k, v|
-          account[k] = Sequel::CURRENT_TIMESTAMP == v ? Time.now : v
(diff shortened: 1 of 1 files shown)


jeremyevans:
Thanks for the patch. This does seem like an obvious bug, and I'll also not sure why the tests didn't catch it.  Maybe because they don't end up using the resulting sms after the call to update_sms?  We should probably add a test that breaks without this change.  I'll see if I can work on that tomorrow, then merge this.

bjeanes:
No worries. I was using `update_hash_ds` in a local feature of mine and noticed the bug so just wanted to bring your attention to it. I would have added a test but I made this PR via GH UI because I'm on a new dev environment without DB setup for rodauth tests. I thought it best to just get your eyes on it first. :)

PR #223, opened 2022-03-05
Update paths.rdoc guide to include use of `require_login_redirect`

jaryl:
When changing `login_route` as prescribed in the guide, developers of non API only apps will likely want to also update `require_login_redirect`. I was looking into whether this was a rodauth-rails specific issue, but I am not familiar enough to tell, and will leave this as draft.

### DIFF
--- a/doc/guides/paths.rdoc
+++ b/doc/guides/paths.rdoc
@@ -8,6 +8,9 @@ corresponding <tt>*_route</tt> method:
 
     # Change login route to "/signin"
     login_route "signin"
+    
+    # Change redirect when login is required to "/signin"
+    require_login_redirect { login_path }
 
     # Change create account route to "/register"
     create_account_route "register"


janko:
I was thinking it would be convenient if the login feature would override `require_login_redirect` with `login_path`, so that it updates automatically.

jaryl:
> I was thinking it would be convenient if the login feature would override require_login_redirect with login_path, so that it updates automatically.

@janko `require_login_redirect` is used in at least 8 other features (going by a GH [search)](https://github.com/jeremyevans/rodauth/search?q=require_login_redirect), so I'd be wary of unintended consequences. I'm fine with having to change this manually (having control is good), but I had to hunt around the docs/code to figure it out.

jeremyevans:
I think this is a good change to make.  I'm also wary of changing the default for `require_login_redirect`, at least until Rodauth 3.0.  I'll merge this shortly.

PR #229, opened 2022-03-24
Change Rails project references into links

ianrandmckenzie:
I also added @janko's project. I thought I'd save my fellow Rails devs researching authentication options a couple of seconds to find information about these projects. With any luck, this PR will collectively save humanity a couple of hours for doing (hopefully) more important things. :)

As you can see from my local repo, I am obviously kind of terrible at this stuff which is why I probably just do safe stuff like READMEs hahaha.

### DIFF
--- a/README.rdoc
+++ b/README.rdoc
@@ -1495,9 +1495,10 @@ required to run the current version of Rodauth is 1.9.2.
 
 All of these are Rails-specific:
 
-* Devise
-* Authlogic
-* Sorcery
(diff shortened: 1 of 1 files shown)


ianrandmckenzie:
Makes sense re: rodauth-rails. It's funny, I normally do links the way you did it, but I followed the conventions of the first links I saw which were links to your website and whatnot. Either way, thanks for taking the idea :)

PR #232, opened 2022-03-25
Extract request halting into `#return_response` method

janko:
This change allows libraries integrating with Rodauth to react on thrown responses by overriding `#return_response` and `#redirect`. It also makes the code a bit DRYer.

I would like to use this in rodauth-rails. Currently, when a Rodauth method that redirects or otherwise returns a response is called from within a Rails controller, the response status is missing from Action Controller's instrumentation output. With this change, rodauth-rails' `rails` feature can override `#redirect` and `#return_response` to save the response status, and assign it for instrumentation.

<details>
<summary>rodauth-rails change</summary>

```diff
diff --git a/lib/rodauth/rails/controller_methods.rb b/lib/rodauth/rails/controller_methods.rb
index 0d80fc0..9114aa3 100644
--- a/lib/rodauth/rails/controller_methods.rb
+++ b/lib/rodauth/rails/controller_methods.rb
@@ -16,6 +16,13 @@ module Rodauth
         request.env.fetch ["rodauth", *name].join(".")
       end
 
+      def append_info_to_payload(payload)
+        super
+        if request.env["rodauth.rails.status"]
+          payload[:status] = request.env["rodauth.rails.status"]
+        end
+      end
+
       private
 
       def rodauth_response
diff --git a/lib/rodauth/rails/feature/instrumentation.rb b/lib/rodauth/rails/feature/instrumentation.rb
index beba6b9..9d86268 100644
--- a/lib/rodauth/rails/feature/instrumentation.rb
+++ b/lib/rodauth/rails/feature/instrumentation.rb
@@ -10,6 +10,14 @@ module Rodauth
 
         def redirect(*)
           rails_instrument_redirection { super }
+        ensure
+          request.env["rodauth.rails.status"] = response.status
+        end
+
+        def return_response(*)
+          super
+        ensure
+          request.env["rodauth.rails.status"] = response.status
         end
 
         def rails_render(*)
```

Before:

```
Started GET "/posts" for ::1 at 2022-03-25 14:16:27 +0100
Processing by PostsController#index as HTML
Redirected to /login
Completed   in 4ms (ActiveRecord: 0.0ms | Allocations: 5568)
```

After:

```
Started GET "/posts" for ::1 at 2022-03-25 14:16:39 +0100
Processing by PostsController#index as HTML
Redirected to /login
Completed 302 Found in 1ms (ActiveRecord: 0.0ms | Allocations: 338)
```
</details>

### DIFF
--- a/lib/rodauth/features/base.rb
+++ b/lib/rodauth/features/base.rb
@@ -511,6 +511,11 @@ module Rodauth
       request.redirect(path)
     end
 
+    def return_response(body=nil)
+      response.write(body) if body
+      request.halt
+    end
+
     def route_path(route, opts={})
       path  = "#{prefix}/#{route}"
       path += "?#{Rack::Utils.build_nested_query(opts)}" unless opts.empty?
--- a/lib/rodauth/features/http_basic_auth.rb
+++ b/lib/rodauth/features/http_basic_auth.rb
@@ -27,7 +27,7 @@ module Rodauth
     def require_http_basic_auth
       unless http_basic_auth
         set_http_basic_auth_error_response
-        request.halt
+        return_response
       end
     end
 
--- a/lib/rodauth/features/json.rb
+++ b/lib/rodauth/features/json.rb
@@ -156,8 +156,7 @@ module Rodauth
         end
       elsif only_json?
         response.status = json_response_error_status
-        response.write non_json_request_error_message
-        request.halt
+        return_response non_json_request_error_message
       end
 
       super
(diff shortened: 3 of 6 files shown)


janko:
Hmm, I actually consider this more of an internal method. If we provided the `return_response` configuration method, I feel it might communicate to the developer that it's called every time Rodauth returns a response, which is not the case (it's not called on redirects & Roda routing matchers).

jeremyevans:
That's a good point.  I'll try testing and merging this later today.

PR #234, opened 2022-04-15
Recommend forwarding response headers with middleware

janko:
Now that Roda 3.55 has been released, which adds the `:forward_response_headers` option to middleware plugin, I thought it would be useful to recommend using this option with Rodauth.

### DIFF
--- a/README.rdoc
+++ b/README.rdoc
@@ -1294,6 +1294,12 @@ By setting <tt>env['rodauth'] = rodauth</tt> in the route block
 inside the middleware, you can easily provide a way for your
 application to call Rodauth methods.
 
+If you're using the remember feature with +extend_remember_deadline?+ set to
+true, you'll want to load roda's middleware plugin with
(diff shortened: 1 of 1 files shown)


jeremyevans:
In the example code you are modifying, no response headers are set.  I don't believe Rodauth itself sets response headers for routes it does not handle.  Are there cases where they actually happens?  Basically, in the example given, does it make sense to use the option?  If not, I don't think the example code should include the option.  I'm fine mentioning the option in text below the example, explaining when it may be useful.

janko:
Good point, no need to load the option if it's not needed. I think the only real use case is the remember feature, so I updated the PR to mention that.

jeremyevans:
This looks good. I'll merge on Monday.

jeremyevans:
Cherry-picked at aca3f01021eb9645170e3c10f9964eafad70c0bd

PR #235, opened 2022-04-18
Fix removing all MFA methods not removing SMS session

janko:
On SMS setup and authentication, `sms_code` is added to the list of methods the session is authenticated by. However, when removing all multifactor authentication methods, Rodauth is removing `sms_codes` method, which will cause the account authenticated via SMS to stay SMS authenticated. We fix that by removing `sms_code` method instead.

The method deletion from session when removing all MFA methods wasn't tested for any MFA method (except implicitly for recovery codes), so we add the missing tests.

### DIFF
--- a/lib/rodauth/features/sms_codes.rb
+++ b/lib/rodauth/features/sms_codes.rb
@@ -468,7 +468,7 @@ module Rodauth
     end
 
     def _two_factor_remove_all_from_session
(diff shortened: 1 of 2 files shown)


jeremyevans:
Thanks for finding and fixing this.  I'll merge later today.

PR #236, opened 2022-04-18
Ensure WebAuthn tests run on Ruby 3.1

janko:
Ruby 3.1 ships with OpenSSL 3.0 as the default gem, but WebAuthn gem don't support OpenSSL 3.0 yet, which caused WebAuthn tests to be skipped. To work around that, we make sure to install OpenSSL 2.x and require it for now.

### DIFF
--- a/.ci.gemfile
+++ b/.ci.gemfile
@@ -86,6 +86,9 @@ platforms :ruby do
   elsif RUBY_VERSION > '2.3'
     gem 'webauthn', '<2.2.0'
   end
+  if RUBY_VERSION > '2.3'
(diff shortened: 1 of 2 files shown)


jeremyevans:
This looks good.  I've tried to get the WebAuthn developers to fix this, since the openssl 3.0 gem works fine for Rodauth's tests, it apparently just doesn't work if you are using the OpenSSL 3 library (I'm using LibreSSL).  I'll merge this later today.

jeremyevans:
Cherry-picked at 4c31c426ca01ab4da430755329f020b03e0e3d0e

PR #238, opened 2022-04-25
Add `#otp_available?` and `#recovery_codes_available?` auth methods

janko:
This is symmetrical to the `#sms_available?` method from the sms_codes feature. These methods should make it easier for the developer to manually generate authentication links, for cases when they cannot use `#two_factor_auth_links`.

In my case, I wanted that requiring multifactor authentication always redirects to the OTP auth page, and that links for authenticating via SMS or recovery code are displayed below the form. This way I would eliminate an additional click needed to get to the recommended MFA page (assuming TOTP is the primary method).

```erb
<!-- otp auth form -->
<% if rodauth.recovery_codes_available? %>
  <%= link_to "Authenticate Using Recovery Code", rodauth.recovery_auth_path %>
<% end %>
<% if rodauth.sms_available? %>
  <%= link_to "Authenticate Using SMS Code", rodauth.sms_request_path, data: { turbo_method: :post } %>
<% end %>
```

While here, I DRYed up `#sms_available?` a bit by having it call `#sms_setup?`.

### DIFF
--- a/doc/otp.rdoc
+++ b/doc/otp.rdoc
@@ -70,6 +70,7 @@ before_otp_setup_route :: Run arbitrary code before handling an OTP authenticati
 otp :: The object used for verifying OTP authentication attempts.
 otp_add_key(secret) :: Add an OTP key for the current account with the given secret.
 otp_auth_view :: The HTML to use for the OTP authentication form.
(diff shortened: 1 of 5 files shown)


jeremyevans:
Thanks for the patch!  This looks good, I'll try merging later today.

PR #241, opened 2022-05-10
Add citext creation to migration in readme

greenfork:
Otherwise it complains that can't find this data type.


jeremyevans:
Needing to load the citext extension is mentioned earlier in the README (http://rodauth.jeremyevans.net/rdoc/files/README_rdoc.html#label-Load+extensions).  It's not in the migration because it requires database superuser permissions, and we don't want to encourage anyone running using a database superuser account for their application (it's a really bad idea from a security perspective).

PR #245, opened 2022-05-30
Ensure routes can be disabled by setting `*_route` to `nil`/`false`

janko:
There are use cases where it makes sense to want to disable certain routes (see https://github.com/janko/rodauth-rails/discussions/110). It appears that setting the `*_route` method to `nil` works, while still keeping the internal request functionality intact. If you agree, I thought it would be useful to have it documented in an official guide.

### DIFF
--- a/doc/guides/paths.rdoc
+++ b/doc/guides/paths.rdoc
@@ -37,3 +37,15 @@ setting:
 
     # ...
   end
+
+There are cases where you may want to disable certain routes. For example, you
+may want to enable the create_account feature to allow creating admins, but
+only make it possible programmatically via internal requests. In this case,
+you should set the corresponding <tt>*_route</tt> method to +nil+:
+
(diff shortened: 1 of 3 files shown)


jeremyevans:
Looks like this works mostly by accident and not by design now, and might still result in the route being available at `/` instead of `/create-account` (didn't test that, but that's how it appears to work).  At the very least, we'd want to make sure we had tests for this behavior, and it would be a good idea to make it deliberate in the code.  I'm guessing the best way to make it deliberate would be to change `post_configure` to not add the route to the route_hash if the route method returns `nil`/`false`.  Is that something you feel comfortable working on?

janko:
Makes total sense 👍🏻. I updated the code to avoid storing `/` into the route hash, tested that both `nil` and `false` can be set, and that internal requests still work.

PR #250, opened 2022-07-05
Add spec for potential security issue

janko:
The owner of the account might to log in to their account via a password on a shared computer, only to realize they don't have access to their MFA device/app, preventing them from continuing. They can then leave the computer, assuming the app handled it as a failed login attempt.

Next time they log in at home, together with multifactor authentication. They might then decide to disable MFA, deciding MFA is not worth it, to prevent situations where they cannot log in on a shared computer. At this point, the shared computer might still have them logged in via a password.

If `#uses_two_factor_authentication?` wasn't cached, another person could now go onto the shared computer and use this account, given that MFA was disabled. However, because it's cached, the session still thinks MFA is setup and the app demands 2nd factor, redirecting them to an empty `/multifactor-auth` page. So, the other person is correctly prevented from accessing the account.

I was first under the impression the scenario was allowed, and that the other person could then hijack the account by setting up MFA, but then learned it's not thanks to `#uses_two_factor_authentication?` being cached. So, I thought I'd add a spec for it 🙂

### DIFF
--- a/spec/two_factor_spec.rb
+++ b/spec/two_factor_spec.rb
@@ -1928,6 +1928,36 @@ describe 'Rodauth OTP feature' do
     end
   end
 
+  it "should prevent authentication when logged in via password and MFA was disabled in another session" do
+    rodauth do
+      enable :login, :otp, :internal_request
+      domain 'example.com'
+    end
+    roda do |r|
+      r.rodauth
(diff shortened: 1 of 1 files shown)


jeremyevans:
Cherry-picked at e6149401414469fb41f1e99063fac88788db204d

PR #265, opened 2022-09-29
Set configuration name to internal request classes

janko:
Since the `@configuration_name` instance variable is the only one not automatically inherited on subclassing, it got lost on the internal request class, which can break user code that rely on it being set. For example, I had code that would save an account type to the database based on the configuration name, but it didn't work correctly when creating the account via an internal request. This PR should fix the issue.

### DIFF
--- a/lib/rodauth/features/internal_request.rb
+++ b/lib/rodauth/features/internal_request.rb
@@ -340,6 +340,7 @@ module Rodauth
 
       klass = self.class
       internal_class = Class.new(klass)
+      internal_class.instance_variable_set(:@configuration_name, klass.configuration_name)
 
       if blocks = klass.instance_variable_get(:@internal_request_configuration_blocks)
         configuration = internal_class.configuration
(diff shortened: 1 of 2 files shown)


lib/rodauth/features/internal_request.rb
```diff
@@ -340,6 +340,7 @@ def post_configure
 
       klass = self.class
       internal_class = Class.new(klass)
+      internal_class.class_eval{@configuration_name = klass.configuration_name}
```
jeremyevans, line 343: Seems better to use `instance_variable_set` here.

spec/rodauth_spec.rb
```diff
@@ -182,32 +182,22 @@
     auth_class.features.must_include :login
   end
 
-  it "should set configuration name for anonymous classes" do
```
jeremyevans, line 185: Can we keep both current specs, and then add the spec you wrote as a new spec?

janko:
Addressed the feedback 👍🏻

jeremyevans:
Cherry-picked at 87d5533a058c8cf3f5672331b4006dae92fec4b7

PR #268, opened 2022-09-30
Extract `account_table_ds` private method

janko:
As discussed in https://github.com/jeremyevans/rodauth/discussions/267, this is useful when sharing database tables between different Rodauth configurations, and wanting to store the account type alongside account records, as it allows filtering retrieved accounts in a single place.

Example usage:

```rb
plugin :rodauth, name: :admin do
  enable :create_account, :login

  before_create_account do
    account[:type] = "admin"
  end

  auth_class_eval do
    def account_table_ds
      super.where(type: "admin")
    end
  end
end
```

### DIFF
--- a/lib/rodauth/features/base.rb
+++ b/lib/rodauth/features/base.rb
@@ -658,7 +658,7 @@ module Rodauth
     end
 
     def _account_from_login(login)
-      ds = db[accounts_table].where(login_column=>login)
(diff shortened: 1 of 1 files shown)


jeremyevans:
Thanks for the patch. This looks good.  I'll try to test and merge today, but it might slip to next week.

PR #269, opened 2022-10-03
Skip displaying links for disabled routes

janko:
I recently added an admin section to the demo Rails app, and I wanted to show the option of enabling the create_account feature, but allowing account creation only through internal requests, by disabling the HTTP route via `create_account_route nil`. This worked, but I noticed that the login page still displayed the registration link, even though the route was disabled. Since there is no information about the route name anyway, I thought it would make sense to skip displaying it.

I then applied this idea to all other places where links are being displayed. I also updated `*_path` and `*_url` methods to return `nil` for disabled routes.

### DIFF
--- a/lib/rodauth.rb
+++ b/lib/rodauth.rb
@@ -126,8 +126,8 @@ module Rodauth
       route_meth = :"#{name}_route"
       auth_value_method route_meth, default
 
-      define_method(:"#{name}_path"){|opts={}| route_path(send(route_meth), opts)}
-      define_method(:"#{name}_url"){|opts={}| route_url(send(route_meth), opts)}
+      define_method(:"#{name}_path"){|opts={}| route_path(send(route_meth), opts) if send(route_meth)}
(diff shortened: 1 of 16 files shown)


jeremyevans:
Thanks for the patch! I agree with the feature in general.  However, in terms of the implementation, I think it would be better to filter the links in a single place to remove `nil` paths, instead of adding/updating conditionals all over the place.  What are your thoughts on that?

janko:
@jeremyevans Makes sense, updated 👍🏻. I wanted to do the filtering in the Rodauth methods itself, rather than in the views, because the links for two_factor_base are used in business logic as well. The only way I saw how to do it was dropping down from `auth_cached_method` to `auth_private_methods`.

I saw a benefit in sorting there, to get consistent order in `/multifactor-{auth,manage,disable}` routes, regardless of the order in which features were loaded. However, given that this can technically break backwards compatibility for applications that are currently relying on that order, I can undo that change if needed.

jeremyevans:
Implementation looks good, thanks!  I'll test and merge later today.

PR #270, opened 2022-10-11
Link rodauth-guest gem on the website

janko:


jeremyevans:
Thanks for the patch!

PR #271, opened 2022-10-20
Raise explicit error when `hmac_secret` is not set but is required

janko:
Some features such as webauthn and active sessions require `hmac_secret` to be set, but raise a cryptic error when it's missing:

```
TypeError: no implicit conversion of nil into String
```

Since it's not clear from this error that the issue is in `hmac_secret` not being set, we change this to raise an explicit `ArgumentError` (like the JWT feature does for `jwt_secret`).

### DIFF
--- a/lib/rodauth/features/base.rb
+++ b/lib/rodauth/features/base.rb
@@ -679,6 +679,8 @@ module Rodauth
     end
 
     def compute_raw_hmac(data)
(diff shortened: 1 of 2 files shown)


jeremyevans:
Thanks for the patch.  I agree this is a good way to fix the issue.

PR #274, opened 2022-10-20
Use Ruby 1.9+ Hash syntax in the README

janko:
This converts hashrockets in the documentation to the Ruby 1.9+ syntax, as discussed in https://github.com/jeremyevans/rodauth/discussions/27 (only README had it).


jeremyevans (commented):

PR #279, opened 2022-11-07
Fix black QR code in TOTP feature

janko:
The `#` is already prepended to the passed fill color, so we need to pass it without a `#`, otherwise we get a `##fff` fill, which results in a completely black square.

Fixes https://github.com/janko/rodauth-rails/issues/169

### DIFF
--- a/lib/rodauth/features/otp.rb
+++ b/lib/rodauth/features/otp.rb
@@ -308,7 +308,7 @@ module Rodauth
     end
 
     def otp_qr_code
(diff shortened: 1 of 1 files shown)


jeremyevans:
Thanks for the patch!  Sorry for the breakage.  I'll try to release 2.26.1 with this fix tomorrow.

I wonder why this resulted in a regression.  You would assume that an invalid fill would be treated as no fill at all, with the background remaining transparent.  That's how invalid CSS rules are handled, but maybe SVG is handled differently.

janko:
Yeah, I was also surprised.

gismoog:
Thanx for the quick fix Jeremy.

PR #288, opened 2022-12-10
Make Rodauth specs pass on Rack 3.0.2+

janko:
Starting from version 3.0.2, `Rack::Utils.build_nested_query` percent-escapes square brackets. See https://github.com/rack/rack/pull/1989.


jeremyevans:
Thanks for the patch!  I'm currently traveling, but I'll merge this after I get back.

PR #289, opened 2022-12-10
Raise error when Sequel database is missing

janko:
Currently, when a person requires Sequel but doesn't connect to the database, they get a `NoMethodError` because `database_type` is getting called on `nil` in `#post_configure` method. I thought it would be good to have a more explicit error message for people newer to Sequel.

### DIFF
--- a/lib/rodauth/features/base.rb
+++ b/lib/rodauth/features/base.rb
@@ -264,7 +264,7 @@ module Rodauth
     end
 
     def db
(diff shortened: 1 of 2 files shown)


jeremyevans:
This makes sense to me, thanks for the patch!  I'll merge this after I return from traveling.

jeremyevans:
Cherry-picked at 0510740768e2b2defa343ea84463f6a8976a8aae

PR #296, opened 2023-01-24
Support configuring symbol param keys for convenience

janko:
People how are used to accessing request parameters via symbol keys (Rails, Sinatra) will probably try configuring `*_param` values with symbol keys, and be surprised the param is not being found (see https://github.com/janko/rodauth-rails/discussions/191). We can have their back by automatically converting set param keys into strings.

I was considering doing this change only in rodauth-rails, but I didn't see how I can handle the internal_request part, given that the `rails` feature has to be loaded before any user configuration. I thought this change would be useful for folks coming from other frameworks as well, such as Sinatra, where it's idiomatic to access params via symbol keys.

### DIFF
--- a/lib/rodauth/features/base.rb
+++ b/lib/rodauth/features/base.rb
@@ -480,7 +480,7 @@ module Rodauth
     end
 
     def raw_param(key)
-      request.params[key]
+      request.params[key.to_s]
     end
 
     def base_url
--- a/lib/rodauth/features/internal_request.rb
+++ b/lib/rodauth/features/internal_request.rb
@@ -47,7 +47,7 @@ module Rodauth
     end
 
     def raw_param(k)
-      @params[k]
+      @params[k.to_s]
     end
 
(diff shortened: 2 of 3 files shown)


jeremyevans:
Sorry, but I don't want to support this by default.  In Roda, param keys are accessed only via strings by default (the indifferent_params plugin is available, but its use is not encouraged).  I'll consider the addition of a `convert_param_key` method that just returns the argument by default, and have `raw_param` use that.  That way people that want the support (or similar support) can use the method to implement what they want.

janko:
Thanks for the input, that makes sense, I will send out a PR that adds what you described 👍

janko:
I ended up adding examples of overriding param keys to the generated Rodauth configuration, e.g. `login_param "email"`, to show the developer they should use string keys. I think that will be good enough, and I get to avoid diverging from default Rodauth behaviour, which I'm generally trying to maintain in rodauth-rails 🙂

PR #298, opened 2023-01-30
Extract `webauthn_key_insert_hash` auth method

janko:
I was live-streaming integrating WebAuthn with Rodauth, where users can additionally set custom names for authenticator, inspired by apps such as RubyGems.org.

I couldn't add the name column value to the insert hash without copying the code, and given that the webauthn credential from form submission isn't accessible in hooks, I also didn't find an easy way to find the inserted record so that I can update it. So, I thought it would be convenient to have an auth method that can be overridden to add additional columns to webauthn keys.

I considered storing the webauthn credential in an instance variable, to avoid having to pass arguments to `webauthn_key_insert_hash`, but saw that it required multiple changes and possibly wasn't desirable, so I chose the more conservative approach with less changes. This is also more consisent with existing methods that accept the webauthn credential argument.


jeremyevans:
Thanks for the patch! This looks fine and makes sense to me.  I'll test and merge shortly.

PR #299, opened 2023-02-06
Make `password_hash` method public

janko:
There are use cases where one might want to create a password hash outside of the request. The rodauth-model gem currently [calls](https://github.com/janko/rodauth-model/blob/9b654f50c284204eda4e92b322b0142a919e91dc/lib/rodauth/model/sequel.rb#L26) this method in the `#password=` setter. The rodauth-rails gem is also now [calling](https://github.com/janko/rodauth-rails/blob/0aeb841b2e9a5950182901c9b3ac2fc8f87bd787/lib/generators/rodauth/templates/test/fixtures/accounts.yml#L4) this method in Active Record fixtures, to more easily support both bcrypt and argon2 along with a potential password pepper.

### DIFF
--- a/lib/rodauth/features/argon2.rb
+++ b/lib/rodauth/features/argon2.rb
@@ -15,6 +15,18 @@ module Rodauth
     auth_value_method :argon2_secret, nil
     auth_value_method :use_argon2?, true
 
+    def password_hash(password)
+      return super unless use_argon2?
+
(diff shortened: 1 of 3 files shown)


jeremyevans:
Thanks for the patch. This seems fine to me.  It wasn't private for a particular reason, just no obvious reason to make it public.  I'll test and merge later today.

PR #302, opened 2023-02-14
Add password maximum length and requirements examples

janko:
I saw in the [OWASP Authentication Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Authentication_Cheat_Sheet.html#implement-proper-password-strength-controls) that setting a maximum password length is recommended to prevent long password Denial of Service attacks, so I thought it's useful to show an example in the guide.

I also recently implemented a custom password complexity validation, and since it required reading the source code, I thought it would be useful to have a concrete example in the docs that people can adapt.


jeremyevans:
Thanks for the patch!

PR #303, opened 2023-02-16
Skip rendering reset password request from on invalid login attempt for internal requests

janko:
The internal request auth subclass doesn't have real request context, and doesn't execute Roda's route block, so it might not have what is necessary to render view templates.

I encountered this when using Rodauth as a library, where the reset password request form failed to render due to missing session handler, raised when Rodauth attempted to generate the CSRF tag.

### DIFF
--- a/lib/rodauth/features/reset_password.rb
+++ b/lib/rodauth/features/reset_password.rb
@@ -219,7 +219,7 @@ module Rodauth
     attr_reader :reset_password_key_value
 
     def after_login_failure
-      unless only_json?
+      unless only_json? || internal_request?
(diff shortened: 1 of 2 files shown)


janko:
I was wondering whether it would make sense for the internal_request mixin to fail when any view attempts to be rendered, so that the developer gets a more explicit error message.

jeremyevans:
Thanks for the patch.  I'll test and merge shortly.

In terms of failing when any views are rendered, you can try that out and see if it breaks any specs. The mailing uses `render` and not `view`, so at least that would have to be allowed.

PR #304, opened 2023-02-16
Add more specs for early rendering on internal requests

janko:
After https://github.com/jeremyevans/rodauth/pull/303, I found more places that are early rendering, which weren't being covered by existing internal request specs (`#before_login_attempt` & `#new_account` in verify_account, `#show_lockout_page` in lockout). Everything seems to be working fine, due to `#set_error_flash` being called before view rendering, but hopefully it's fine to add specs to ensure it continues working 🙂


jeremyevans:
I agree, it's good to have these to avoid future regressions.  Thanks for the patch!

PR #305, opened 2023-02-18
Show adding error reason to the JSON response

janko:
This shows how the JSON response can be modified in general, which I believe isn't currently mentioned in the docs. It also makes error reasons functionality easier to discover; they're currently only mentioned in base feature docs, but they'll most likely be used for JSON APIs anyway, so I think it's useful to mention them there 🙂


jeremyevans:

PR #314, opened 2023-03-04
Fix duplicate active sessions when creating and verifying account

janko:
When verify_account_grace_period feature is logged in, when both `create_account_autologin?` and `verify_account_autologin?` are set to `true`. This means that when creating account and then verifying it, `#autologin_session` will be called twice. Since an active session is created on each `#update_session` call, the user will end up with two active sessions, with the second one being the current one. We fix this by removing the existing current session (if it exists) before proceeding to update the session.

### DIFF
--- a/lib/rodauth/features/active_sessions.rb
+++ b/lib/rodauth/features/active_sessions.rb
@@ -101,6 +101,7 @@ module Rodauth
     end
 
     def update_session
+      remove_current_session
       super
       add_active_session
     end
--- a/spec/active_sessions_spec.rb
+++ b/spec/active_sessions_spec.rb
@@ -72,6 +72,30 @@ describe 'Rodauth active sessions feature' do
     end
   end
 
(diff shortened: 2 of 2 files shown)


jeremyevans:
Sure, this looks fine.  Just to make sure I understand the reasoning behind the change, the purpose of this patch is to clean up the session early, correct (it would timeout eventually without this change)?  Note that if the verify account happens in a different browser than the create account, then you'll still have two sessions (this is expected when using active_sessions).

janko:
Yes, it's to clean the session early. I have `session_inactivity_deadline` set to 14 days, so in my case it would take longer for the session to expire. But it's also possible to set both `session_inactivity_deadline` and `session_liftetime_deadline` to `nil`, in which case I believe the duplicated session would never expire by itself (unless logging out of all sessions). And yeah, I expect to have two sessions when creating account in one browser and verifying it in the other 👍🏻

PR #317, opened 2023-03-07
Extract `remove_active_session(session_id)` method

janko:
This is useful when implementing session revoking, where we want to delete a specific session from the database that's not the current session, in order to log that browser out of the app.

This requires a HMAC'ed session ID, because it's intended to be called using a session ID retrieved from the database, which is HMAC'ed.


jeremyevans:
I'm OK adding this. By default, the method is isn't all that useful, as Rodauth doesn't offer any way to distinguish active sessions.  I assume you are keeping some other metadata related to the active session (IP/User-Agent/etc.), that you can use to help determine which session to remove?

janko:
Thank you for merging 🙏🏻 Yes, I'm storing IP address, User-Agent, and location on the session record as well, and presenting this information similar to how GitHub does.

PR #319, opened 2023-03-13
Support :render=>false plugin option

davekaro:
This should be used if alternative view libraries will be provided. For example, if using Rodauth with Phlex to render HTML.


jeremyevans:

PR #321, opened 2023-03-18
Extract setting remember cookie into a separate method

janko:
This should make the `remember_login` method a bit easier to follow, because setting the remember cookie is a bit more involved, so for me it's beneficial to name that unit of work. We also do it for `forget_login` for consistency.

### DIFF
--- a/lib/rodauth/features/remember.rb
+++ b/lib/rodauth/features/remember.rb
@@ -127,13 +127,7 @@ module Rodauth
 
     def remember_login
       get_remember_key
-      opts = Hash[remember_cookie_options]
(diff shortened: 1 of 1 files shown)


jeremyevans (commented):
I'm not generally in favor of moving code called a single place into a method, unless there is a desire to override the behavior. Especially in `forget_login` where all code is being moved into another method.  Do you plan on overriding these new methods?  If not, I would be OK with the `set_remember_cookie` extraction, but I don't see the point of moving all `forget_login` code into another method and having `forget_login` call that.

janko:
@jeremyevans Makes sense, updated the PR 👍🏻

PR #322, opened 2023-03-21
Remove unmatched </div>

fdr:
I was trying to run this through an erb auto formatter, and it crashed on this. Looking closely, emacs also noticed the unmatched `</div>`. Possibly the indentation in the surrounding code is a little weird and could be changed.


jeremyevans:
Thanks for the patch!

PR #330, opened 2023-04-21
Replaced Array#pack with Base64.urlsafe_encode64 in specs for generating authorization header

igor-alexandrov:
Current implementation of Authorization header in specs may produce not URL-safe. I found this in [rodauth-oauth](https://gitlab.com/os85/rodauth-oauth/-/merge_requests/121), but this also applies to Rodauth.

### DIFF
--- a/spec/http_basic_auth_spec.rb
+++ b/spec/http_basic_auth_spec.rb
@@ -7,7 +7,7 @@ describe "Rodauth http basic auth feature" do
   end
 
   def authorization_header(opts={})
-    ["#{opts.delete(:username)||'foo@example.com'}:#{opts.delete(:password)||'0123456789'}"].pack("m*")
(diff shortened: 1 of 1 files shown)


jeremyevans:
Thanks for submitting a pull request. Can you explain why we should make this change?  This is in the specs and not in the library itself, so it should have no effect on Rodauth usage.

Even if this was in the library, the Authorization header is not designed to be used in URLs.  Is there a reason it matters that the header values are not URL safe?  If you want to use header values in URLs, you should URL encode them.

HoneyryderChuck:
@jeremyevans while I agree that this is test code and correctness enforcement is not a must, the current implementation is wrong, as it allows newlines, which break the regex used in the production code to parse the token. I agree that making is urlsafe is probably too much (encode64 would have sufficed), it should at least be changed to"pack('m0')" in order to be correct.

jeremyevans:
@HoneyryderChuck Thanks for explaining that.  I'll switch to `m0`.

PR #340, opened 2023-05-26
Added new auth method `json_response_error?`

opya:
Based on this discussion https://github.com/jeremyevans/rodauth/discussions/339 and @jeremyevans suggestions.


jeremyevans:
I merged the first patch but not the second. The second patch doesn't fix a typo, the email address is deliberately wrong (even the domain is wrong).

opya:
Thank you! Jeremy

PR #347, opened 2023-07-05
Mention active_sessions in JWT feature docs

janko:
It's already mentioned in JWT Refresh docs, but it also works just with JWT, so I thought it's good to mention it there as well. We clarify that this implements the whitelist approach to revoking JWT tokens.


jeremyevans:
Thanks for the patch!

PR #350, opened 2023-07-17
Added Gemfile

igor-alexandrov:
## Description
This pull request adds a Gemfile to the Rodauth project. By including a Gemfile, it becomes easier for users and contributors to manage and install the required dependencies when working locally with Rodauth.

## Changes
 * Added a Gemfile to the project root.
 * Update .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@
 /coverage
 /www/public/*.html
 /www/public/rdoc/
+
+Gemfile.lock
--- a/Gemfile
+++ b/Gemfile
@@ -1,0 +1,4 @@
+source 'https://rubygems.org'
+
+# Specify your gem's dependencies in rodauth.gemspec
+gemspec


janko:
Jeremy doesn't encourage use of Gemfile in libraries, see https://github.com/jeremyevans/sequel/issues/1809 😉

igor-alexandrov:
I see... in this case maybe Contributing guide should be updated. By the way, before creating this PR, I checked that both Roda (https://github.com/jeremyevans/roda/blob/master/Gemfile) and Sequel (https://github.com/jeremyevans/sequel/blob/master/Gemfile) have Gemfile in the repo.

janko:
Oh, somehow I missed that, and was sure I didn't see Gemfiles in other repos, but I guess it was just this one 👍🏻

igor-alexandrov:
So you suggest using `gem install --development rodauth`? Maybe this should be added to contributing?

jeremyevans:
I'll add that to the `Dependencies` section of the `README`.

PR #351, opened 2023-07-17
Added rackup dependency for Rack 3 compatibility

igor-alexandrov:
## Description

In Rack 3 `bin/rackup` (that is used to serve local version of documentation) has been moved to a separate gem: https://github.com/rack/rack/blob/main/UPGRADE-GUIDE.md#binrackup-rackserver-rackhandlerand--racklobster-were-moved-to-a-separate-gem.

### DIFF
--- a/rodauth.gemspec
+++ b/rodauth.gemspec
@@ -49,6 +49,7 @@ END
   s.add_dependency('roda', [">= 2.6.0"])
   s.add_development_dependency('tilt')
   s.add_development_dependency('rack_csrf')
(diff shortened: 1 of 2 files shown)


jeremyevans:
`rackup` isn't required.  You can use any webserver that supports rack applications.

jeremyevans:
Ah, that makes sense.  I think in this case, we are better off removing the `serve` task and `www/config.ru`, as the files are just static files and all links are relative.  I'll do that.

PR #353, opened 2023-07-18
Allow explicit usage of `p_cost` in argon2 feature.

estebanz01:
Based on a discussion on #349 where I ran into a small issue when a custom hash cost for argon2 was specified, this PR implements the explicit usage of the `p_cost` attribute that can be defined when hashing passwords with this particular algorithm.

I didn't run any linter on my changes as I'm not sure which one is used or if there's some specific configuration I need to have.

All test passed on postgresql and ruby 3.2.2, but I can try other versions/databases if needed.

### DIFF
--- a/lib/rodauth/features/argon2.rb
+++ b/lib/rodauth/features/argon2.rb
@@ -43,7 +43,7 @@ module Rodauth
 
     def password_hash_cost
       return super unless use_argon2?
-      argon2_hash_cost 
+      argon2_hash_cost
     end
 
     def password_hash_match?(hash, password)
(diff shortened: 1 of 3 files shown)


estebanz01:
@jeremyevans not sure if you get a notification for open PRs, so I'm going to ping you here 😅 sorry for any duplicated alerts.

jeremyevans:
Looks good, thanks for the patch!  I'll do some testing locally and merge if there are no problems.

I do get notified about open PRs, otherwise it would be difficult to ensure that all PRs are addressed quickly.

jeremyevans:
Squash merged at 49ffc7ee22fd95e4b3dfa086e2629ca3666f8de9

PR #354, opened 2023-07-19
Allow WebAuthn login to count for two factors

janko:
From the [FIDO Alliance website](allow-webauthn-to-count-for-two-factors):

> **Are passkeys considered multi-factor authentication?**
>
> Passkeys are kept on a user’s devices (something the user “has”) and — if the RP requests User Verification — can only be exercised by the user with a biometric or PIN (something the user “is” or ”knows”). Thus, authentication with passkeys embodies the core principle of multi-factor security.

To allow passwordless login via a passkey to be counted for two factors, we add a new `webauthn_login_two_factors?` configuration option, which when set to true will consider the user two-factor authenticated if they're using multifactor authentication.

This is useful for skipping classic 2nd factor methods such as TOTP, recovery codes or SMS codes, but it's especially useful when the account has a password or email_auth is enabled, both of which are considered additional factors, and would normally be required after WebAuthn login.

In this case, `webauthn_user_verification` should probably be set to `preferred` (set by `webauthn_autofill` feature) or `required`, rather than `discouraged` set by `webauthn` feature. Though, even with `discouraged` user verification, my Mac and iPhone still ask for biometric identification.

### DIFF
--- a/doc/webauthn_login.rdoc
+++ b/doc/webauthn_login.rdoc
@@ -5,6 +5,7 @@ WebAuthn. It depends on the login and webauthn features.
 
 == Auth Value Methods
 
+webauthn_login_user_verification_additional_factor? :: Whether passwordless login via WebAuthn should consider user verification as 2nd factor when using multifactor authentication, false by default.
 webauthn_login_error_flash :: The flash error to show if there is a failure during passwordless login via WebAuthn.
 webauthn_login_failure_redirect :: Whether to redirect if there is a failure during passwordless login via WebAuthn.
 webauthn_login_route :: The route to the webauthn login action.
--- a/lib/rodauth/features/webauthn_login.rb
+++ b/lib/rodauth/features/webauthn_login.rb
@@ -10,6 +10,8 @@ module Rodauth
 
     error_flash "There was an error authenticating via WebAuthn"
 
+    auth_value_method :webauthn_login_user_verification_additional_factor?, false
+
     route(:webauthn_login) do |r|
       check_already_logged_in
       before_webauthn_login_route
@@ -24,6 +26,9 @@ module Rodauth
           before_webauthn_login
           login('webauthn') do
             webauthn_update_session(webauthn_credential.id)
+            if webauthn_login_verification_factor?(webauthn_credential)
+              two_factor_update_session('webauthn-verification')
+            end
           end
         end
 
@@ -48,11 +53,22 @@ module Rodauth
       end
     end
 
+    def webauthn_user_verification
+      return 'preferred' if webauthn_login_user_verification_additional_factor?
+      super
+    end
+
     def use_multi_phase_login?
       true
     end
 
     private
+
+    def webauthn_login_verification_factor?(webauthn_credential)
+      webauthn_login_user_verification_additional_factor? &&
+        webauthn_credential.response.authenticator_data.user_verified? &&
+        uses_two_factor_authentication?
+    end
 
     def account_from_webauthn_login
       account_from_login(param(login_param))
--- a/spec/webauthn_login_spec.rb
+++ b/spec/webauthn_login_spec.rb
@@ -203,6 +203,72 @@ describe 'Rodauth webauthn_login feature' do
     page.html.must_include 'Logged In via password and webauthn'
   end
 
+  it "should allow treating user verification as 2nd factor" do
+    rodauth do
+      enable :webauthn_login, :logout
+      hmac_secret '123'
+      webauthn_login_user_verification_additional_factor? true
+      account_password_hash_column :ph
+    end
+    first_request = nil
+    roda do |r|
+      first_request ||= r
+      r.rodauth
+      rodauth.require_authentication
+      r.root{view :content=>"Authenticated by: #{rodauth.authenticated_by}"}
+    end
+
+    visit '/'
+
+    origin = first_request.base_url
+    webauthn_client = WebAuthn::FakeClient.new(origin)
+
+    visit '/login'
+    fill_in 'Login', :with=>'foo@example.com'
+    click_button 'Login'
+    fill_in 'Password', :with=>'0123456789'
+    click_button 'Login'
+
+    visit '/webauthn-setup'
+    challenge = JSON.parse(page.find('#webauthn-setup-form')['data-credential-options'])['challenge']
+    fill_in 'Password', :with=>'0123456789'
+    fill_in 'webauthn_setup', :with=>webauthn_client.create(challenge: challenge).to_json
+    click_button 'Setup WebAuthn Authentication'
+
+    logout
+
+    visit '/login'
+    fill_in 'Login', :with=>'foo@example.com'
+    click_button 'Login'
+    challenge = JSON.parse(page.find('#webauthn-auth-form')['data-credential-options'])['challenge']
+    fill_in 'webauthn_auth', :with=>webauthn_client.get(challenge: challenge).to_json
+    click_button 'Authenticate Using WebAuthn'
+    find("#error_flash").text.must_equal "You need to authenticate via an additional factor before continuing"
+
+    logout
+
+    visit '/login'
+    fill_in 'Login', :with=>'foo@example.com'
+    click_button 'Login'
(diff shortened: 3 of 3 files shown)


janko:
Thanks for the suggestion. I first assumed that the server doesn't know whether user verification took place, but then I found out that this flag *is* actually included in the assertion response. So, I changed the behavior to only add the additional factor when user verification took place, and named it `webauthn-verification` as you suggested.

janko:
I also modified `webauthn_user_verification` to return `preferred` when this setting has been turned on. When I use this feature, I want as many users to avoid having to authenticate with 2nd factor, so I would prefer user verification to happen.

janko:
I guess `preferred` should be set only for passwordless login scenario, and `discouraged` kept when authenticating via WebAuthn as 2nd factor. Probably the same for `webauthn_autofill` feature.

jeremyevans:
This looks good.  I'll test and merge tomorrow.  Thanks for the patch!

janko:
Do you think this behavior could become the default in Rodauth 3? I think most apps would want this, as it reduces the amount of steps for the end user, and my impression was that it's how passkeys were intended to be used (i.e. acting as both factors). The main reason I disabled it by default was caution of backwards compatibility.

jeremyevans:
Making this the default behavior in Rodauth 3 seems reasonable to me.

jeremyevans:
> Making this the default behavior in Rodauth 3 seems reasonable to me.

After thinking about this more, I'm not sure I want to do this by default.  

Consider the common password + webauthn compared to webauthn + user verification.  In the password + webauthn case, you can at least somewhat control password strength (length, complexity, blacklists).  In the webauthn + user verification case, it seems like you have no control over the strength of the user verification.

It doesn't seem like you can enforce the strength of the user verification unless you are controlling and configuring the passkey (i.e. you provide the passkey to the user, instead of the user providing their own). There are definitely environments where this is the case, as I have personal experience with.  However, that doesn't seem to be the common case.

In addition to not being able to control the strength of the user verification, if the user can provide their own authenticator, they could select one that says it is user verified when it is not.  There does not appear to be a way for the server to enforce that the authenticator is telling the truth regarding user verification.  I would guess that commonly available authenticators wouldn't lie about this, but there doesn't seem to be a way to enforce it.

It seems like we would want to document that you should only enable this option if you are providing the authenticators used and are ensuring that they correctly implement user verification to a strength that you accept as an additional factor.

Am I misunderstanding the trust issue here?  Are the benefits of ignoring the issue and trusting by default worth the cost of allowing users to pick authenticators that support weak user verification or lie about user verification?  What are your thoughts?

Note that everything else looks good to merge.  I would just like your feedback before updating the documentation and deciding whether to make this the default behavior in Rodauth 3.

janko:
I appreciate the discussion. As I see it, it's ultimately the user's responsibility to protect their own account. If they chose an authenticator because it lies about user verification, that's on them, they will have a less protected account. Though if they're part of an organization that requires MFA for the security of the organization, then a breach of their account wouldn't hurt only them.

On the other hand, if the user verification is just weak (I'm imagining a 4-digit PIN), the user might trust the app to know whether that is enough to count for two factors. Most users don't really feel like setting up and using 2nd factor, the app usually needs to motivate them. So, in this case I see it as the responsibility of the app to determine what is good enough.

If these are real concerns, then I'm just surprised how come big companies like Google and GitHub treat my passkeys as two factors. My authenticators are my Mac and iPhone, both of which use biometric user verification, but I think they cannot know that. I haven't tested how they treat security keys that have no user verification, as I don't have such an authenticator with me. It helps me to look at what big companies are doing in terms of authentication when I'm trying to figure out what is secure enough.

I also noticed that when I register a passkey on my Google Account, it seems to know where it is stored (iCloud Keychain, 1Password etc). I looked at the list of WebAuthn extensions on MDN, and I couldn't find a way to obtain this information. The most I could find is that the Relying Party can request the [minimum PIN length](https://developer.mozilla.org/en-US/docs/Web/API/Web_Authentication_API/WebAuthn_extensions#minpinlength) from the authenticator, which I presume can then be used to determine whether that's strong enough to count for 2nd factor. I'm now also thinking whether the new configuration option shouldn't also accept the credential as the argument, if there are ways to assess the strength of user verification 🤔

jeremyevans:
> I appreciate the discussion. As I see it, it's ultimately the user's responsibility to protect their own account. If they chose an authenticator because it lies about user verification, that's on them, they will have a less protected account. Though if they're part of an organization that requires MFA for the security of the organization, then a breach of their account wouldn't hurt only them.

For cases where two factor authentication is optional for the user, I agree completely that treating passkey + user verification as two factors makes sense.  For cases where the organization requires MFA for all users, this seems like a simple way to combine two factors into a single factor unless the organization controls the authenticators.

> On the other hand, if the user verification is just weak (I'm imagining a 4-digit PIN), the user might trust the app to know whether that is enough to count for two factors. Most users don't really feel like setting up and using 2nd factor, the app usually needs to motivate them. So, in this case I see it as the responsibility of the app to determine what is good enough.

Maybe there is a way for the authenticator to provide that information when setting up the authentication, but as there is no way for the app to verify it, this again relies on the app trusting the authenticator, which should only be done if the organization running the app controls the authenticators.
 
> If these are real concerns, then I'm just surprised how come big companies like Google and GitHub treat my passkeys as two factors. My authenticators are my Mac and iPhone, both of which use biometric user verification, but I think they cannot know that. I haven't tested how they treat security keys that have no user verification, as I don't have such an authenticator with me. It helps me to look at what big companies are doing in terms of authentication when I'm trying to figure out what is secure enough.

Google and GitHub both are opt-in for MFA and do not require MFA as far as I know, so them treating the passkey + user verification as two factors seems reasonable.

> I also noticed that when I register a passkey on my Google Account, it seems to know where it is stored (iCloud Keychain, 1Password etc). I looked at the list of WebAuthn extensions on MDN, and I couldn't find a way to obtain this information. The most I could find is that the Relying Party can request the [minimum PIN length](https://developer.mozilla.org/en-US/docs/Web/API/Web_Authentication_API/WebAuthn_extensions#minpinlength) from the authenticator, which I presume can then be used to determine whether that's strong enough to count for 2nd factor. I'm now also thinking whether the new configuration option shouldn't also accept the credential as the argument, if there are ways to assess the strength of user verification 🤔

The issue with minPinLength is that it again relies on the app trusting the authenticator.

Thanks for the discussion.  I haven't decided yet on whether to make this the default, but I'll update the documentation regarding this when I merge.

bjeanes:
> I also noticed that when I register a passkey on my Google Account, it seems to know where it is stored (iCloud Keychain, 1Password etc). I looked at the list of WebAuthn extensions on MDN, and I couldn't find a way to obtain this information.

I don't know if an answer to this was found, but I am doing a bit of research on this to try to give Webauthn credentials default names (and, ultimately, allow the user to specify a custom one). I got a proof-of-concept working by using https://github.com/passkeydeveloper/passkey-authenticator-aaguids and a feature:


```rb
module Rodauth
  Feature.define(:webauthn_metadata, :WebauthnMetadata) do
    depends :webauthn

    auth_value_method :webauthn_setup_name_param, "webauthn_key_name"

    def webauthn_attestation
      "indirect" # https://www.w3.org/TR/webauthn-2/#dom-attestationconveyancepreference-indirect
    end

    def webauthn_key_insert_hash(webauthn_credential)
      super.merge({
        name: param(webauthn_setup_name_param).presence,
        aaguid: webauthn_credential.response.aaguid || "00000000-0000-0000-0000-000000000000"
        # attestation_payload: webauthn_credential.response.send(:attestation_object_bytes)
      })
    end

    def account_webauthn_keys
      webauthn_keys_ds.all.map do |key|
        meta = WebauthnMetadata::AAGUIDS[key[:aaguid]] || {"name" => "Unknown"}
        {**meta, **key, default_name: meta["name"]}.with_indifferent_access
      end
    end
  end

  # Maps between AAGUIDs and a map that includes keys for 'name', 'icon_light', 'icon_dark'
  WebauthnMetadata::AAGUIDS = JSON.parse(File.read("/webauthn/webauthn_aaguids.json"))
end
```

The real feature would probably want to use the official metadata service, as that link suggests, but it seems to me that the key to answering questions about the security key or passkey is via inspecting the `webauthn_credential.response`.

PR #355, opened 2023-07-25
Add internal request support for WebAuthn

janko:
The Rails app in my current company uses custom authentication, and for a hack week I wanted to add support for passkeys. For this I wanted to try leveraging Rodauth's implementation, by using Rodauth as a library.

This PR extends the internal_request feature with support for WebAuthn actions. Initially, the `*_params` methods were returning webauthn-ruby's credential objects, but that required manually HMAC'ing the challenge. So, I used the same approach as `otp_setup_params` and returned the data as a hash. I like that this encapsulates webauthn-ruby usage within Rodauth, and the caller just deals with plain hashes.

The webauthn feature adds `webauthn_setup_params`, `webauthn_setup`, `webauthn_auth_params`, `webauthn_auth` and `webauthn_remove` methods, the webauthn_login feature adds `webauthn_login_params` and `webauthn_login` methods, while the webauthn_autofill feature makes `webauthn_login_params` not require the login param. The `webauthn_login` method is the only one that returns something (account ID); I was deciding whether I should make `webauthn_setup` return the credential ID, but realized the caller can already retrieve it from the `navigator.credentials.create` result.

I tested this in a sample app, and everything seems to be working correctly 🙂

### DIFF
--- a/doc/internal_request.rdoc
+++ b/doc/internal_request.rdoc
@@ -461,3 +461,79 @@ account login, before the change.
 
 Options:
 +:verify_login_change_key+ :: The verify login change key for the account. This allows verifying login changes by key, without knowing the account id or login.
+
+=== WebAuthn
+
+==== webauthn_setup_params (requires account)
+
+The +webauthn_setup_params+ method returns a hash with +:webauthn_setup+,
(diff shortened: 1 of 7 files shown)


jeremyevans:
This looks good, thanks for working on this.  I'll try to test and merge later this week.

PR #356, opened 2023-07-25
Throw different field error when WebAuthn credential could not be found in webauthn_autofill

janko:
When using the WebAuthn autofill UI, the `/webauthn-login` route will receive the selected credential without the login param. It's possible the credential was deleted from the database, this will currently fall through to a "no matching login" field error, which isn't accurate because we're not matching by login here.

To improve this, throw a dedicated field error in this case. It won't be visible in HTML, but it helps make the failure clear for JSON and internal requests. I noticed this when I was working on internal request support for WebAuthn, where the current error message was misleading.


jeremyevans:
I agree, it makes sense to use a different error here. Thanks for the patch!  I'll try to test and merge later this week.

PR #357, opened 2023-07-30
Mention that webauthn_autofill allows WebAuthn login without an email

janko:
Generating WebAuthn authentication options without the account's credentials whitelisted is a WebAuthn feature that's not tied to the autofill UI, as I recently learned, it just requires discoverable credentials.

This allows the developer to provide a button for logging in via a passkey, which should have wider browser support compared to conditional mediation. I thought it's useful to mention that in the documentation.


jeremyevans:
Makes sense, thanks for the patch!

PR #361, opened 2023-08-16
Add guide for rendering confirmation view

janko:
*Follow-up on https://github.com/jeremyevans/rodauth/discussions/360*

This is a useful technique for rendering views after successful actions, while avoiding having to define routes for those pages, which would make them accessible outside of those actions.


jeremyevans:

PR #366, opened 2023-09-24
Fix code tag in README

janko:
The `+` symbols are being rendered literally, so probably because of the space we need to use the `<tt>` tag.


jeremyevans:
Thanks for the patch.  RDoc doesn't support spaces inside `+` tags, that's why it doesn't work.

PR #370, opened 2023-10-22
Set flash notifications if message is not nil

igor-alexandrov:
Currently, there is no way to disable flash notifications completely. The example below will still set flash keys with `nil` message. I added a check for nil in methods that set flash notifications. If a message is not present, flash will not be set.

``` ruby
# frozen_string_literal: true

class RodauthImpersonate < Rodauth::Rails::Auth
  configure do
    enable :login, :logout

    prefix '/impersonate'

    session_key_prefix 'impersonate_'

    logout_redirect '/'
    logout_notice_flash nil

    login_redirect nil
    login_route nil
    login_notice_flash nil
  end

  def clear_session
    session.keys.each do |key|
      session.delete(key) if key.start_with?(session_key_prefix)
    end
  end
end

```

### DIFF
--- a/lib/rodauth/features/base.rb
+++ b/lib/rodauth/features/base.rb
@@ -60,7 +60,7 @@ module Rodauth
     auth_value_method :mark_input_fields_with_inputmode?, true
     auth_value_method :skip_status_checks?, true
     auth_value_method :template_opts, {}.freeze
-    auth_value_method :title_instance_variable, nil 
+    auth_value_method :title_instance_variable, nil
     auth_value_method :token_separator, "_"
     auth_value_method :unmatched_field_error_status, 422
     auth_value_method :unopen_account_error_status, 403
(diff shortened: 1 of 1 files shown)


jeremyevans:
I'm OK with this, but you'll need to not make unnecessary whitespace notifications, and you'll need to add covering tests (Rodauth requires 100% line and branch coverage).  That's going to be at least four separate tests (or test assertions), one for each of the cases where you are adding a conditional.

igor-alexandrov:
@jeremyevans sure, will do! I have some problems running tests locally. How it is better to show them to you?

jeremyevans:
You can post the issues you are having with running the tests here.  Note that the repository has CI, so if you can't get the tests working locally, you can use the CI to check.

For making sure the tests are covering, temporarily add `raise unless message` in each of the four methods where you add the conditional, and write tests that will hit that and raise the RuntimeError.  Then you can remove the code and have the test check that no flash is set (not even an empty flash).

PR #375, opened 2023-12-03
Return error response instead of 404 when token is missing

janko:
When a page requiring a token is requested without a token, Rodauth currently returns a 404 response. If Rodauth is used as a middleware with `:next_if_not_found` plugin option enabled, this request would be forwarded to the main app. This behavior can be misleading for the developer, because it might communicate that it's not a route handled by Rodauth.

I encountered this when I mistakenly generated a `/reset-password` link instead of `/reset-password-request`, and then being confused by the response, only to realize what happened because I was familiar with Rodauth internals.

I think it's better to instead return an error response that the required token is missing. This makes things clearer to the developer, and also ensures the request doesn't get forwarded to the main app.

### DIFF
--- a/lib/rodauth/features/email_auth.rb
+++ b/lib/rodauth/features/email_auth.rb
@@ -77,14 +77,12 @@ module Rodauth
           redirect(r.path)
         end
 
-        if key = session[email_auth_session_key]
-          if account_from_email_auth_key(key)
-            email_auth_view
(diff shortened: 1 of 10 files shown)


jeremyevans:
Thanks for the patch.  I agree that this approach makes more sense.  I'll merge and test tomorrow.

Issue #376, opened 2023-12-13
It is possible to verify a phone number for SMS verification without ever verifying it


Bertg:
We have discovered a way that Rodauth allows SMS setup to happen with any phone number, without verification.
The user story is something like this: (this is the issue we confirmed, it may be more wide spread).

```
Given I'm a user and I already have OTP codes set up
When I choose to start the setup of SMS based two factor
But I get interrupted before I can confirm with the code
And, later, I sign into my account again with OTP
Then the SMS setup is confirmed
```

The issue here is the `sms_failures_column` does two things. 

* It keeps track of failures 
* It keeps track of the fact that a code is in setup or not. See [`sms_needs_confirmation?`](https://github.com/jeremyevans/rodauth/blob/cfbd016b5ad06ba44807d382ef0d4bffca023938/lib/rodauth/features/sms_codes.rb#L436).

In itself that is not an issue. However if we look at [`two_factor_authenticate`](https://github.com/jeremyevans/rodauth/blob/cfbd016b5ad06ba44807d382ef0d4bffca023938/lib/rodauth/features/two_factor_base.rb#L248) we see a call to `two_factor_remove_auth_failures`. This method is implemented by all OTP methods, including [the SMS implementation](https://github.com/jeremyevans/rodauth/blob/cfbd016b5ad06ba44807d382ef0d4bffca023938/lib/rodauth/features/sms_codes.rb#L369-L371):

```ruby
    def sms_remove_failures
      update_sms(sms_failures_column => 0, sms_code_column => nil)
    end
```

By setting the `sms_failures_column` to `0` the application has now, by accident, confirmed the phone number. 

In hindsight there is also a code smell in the [`sms_confirm` method](https://github.com/jeremyevans/rodauth/blob/cfbd016b5ad06ba44807d382ef0d4bffca023938/lib/rodauth/features/sms_codes.rb#L373C9-L373C20).

```ruby
    def sms_confirm
      sms_remove_failures
      super if defined?(super)
    end
```

Removing failures to confirm a phone number is one thing, but noting else happens. This code literally reads that confirming sms and removing failures is the same. And it really shouldn't be.

A quick solution would be to make the `sms_remove_failures` method check if `sms_failures_column` is not null.
And write a separate method to write the `sms_confirm` to the database. I think it's best to keep these 2 codepaths separated.

The cleanest way would be to have a non ambiguous way of storing what rows in the `sms_codes_table` are for verification and which ones are for authentication.

[...]

PR #377, opened 2023-12-13
Don't confirm SMS setup during sign in

Bertg:
This is a possible simple fix for #376

- Introduce failing test
- Fix issue where we accidentally accept a non-confirmed SMS code during sign in


jeremyevans:
Thanks for this.  The test is especially helpful.  It passes both with the fix here and the fix I posted in https://github.com/jeremyevans/rodauth/issues/376#issuecomment-1854136723.  I'm thinking of combining the two, that way we can skip the query in `sms_remove_failures` if the sms still needs confirmation.

Bertg:
@jeremyevans Updated PR is here ☝️

PR #381, opened 2023-12-22
Allow retrieving current route name

janko:
As discussed in https://github.com/jeremyevans/rodauth/discussions/380, this is useful for generating path or URL for current route with different parameters, while still reusing configured route definitions.

```rb
# locale switcher with query parameter
rodauth.send(:"#{rodauth.route}_path", request.GET.merge("locale" => "br"))
#=> "/create-account?locale=br"

# locale switcher with path prefix, assuming `prefix { "/#{I18n.locale}/auth" }`
I18n.with_locale("br") { rodauth.send(:"#{rodauth.route}_path", request.GET) }
#=> "/br/auth/create-account"
```

This is the minimal functionality needed to support this use case. I was considering also adding `current_path` and `current_url` methods that would utilize this, but thought that might be an overkill for now.

### DIFF
--- a/lib/rodauth.rb
+++ b/lib/rodauth.rb
@@ -138,6 +138,7 @@ module Rodauth
 
       define_method(handle_meth) do
         request.is send(route_meth) do
+          @route = name
           check_csrf if check_csrf?
           _around_rodauth do
             before_rodauth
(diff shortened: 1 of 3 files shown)


jeremyevans:
Thanks for the patch.  I'll probably have time to review on December 27.

jeremyevans:
Just FYI, I renamed `route` to `current_route` after merging, as I think that is a better name (especially given your spec description).

janko:
Thanks for merging, I agree with the rename 👍🏻

PR #382, opened 2023-12-25
Add Ubicloud to applications using Rodauth

janko:
Judging from https://github.com/ubicloud/ubicloud/pull/952, it seems that they're using Rodauth at an advanced level too.


jeremyevans:
Thanks for the patch!

jeremyevans:
Cherry-picked at 3b4811482fa26ed5bce6fb0926cb98a67635743b

fdr:
It's Good Software

PR #390, opened 2024-01-11
Handle deleted account when checking whether MFA is setup

janko:
It's possible that an account is signed in but the record gets deleted in the database, especially in development where it's common to clear testing data. When `#uses_two_factor_authentication?` is called, it ends up calling `#has_password?`, which errors if the logged in account record doesn't exist. I thought it would be nice if Rodauth gracefully handled that.

I also made two spec changes on the way: one was to use the `PASSWORD_HASH_TABLE` constant, the other was renaming the spec name to communicate that it's testing all MFA behavior.

### DIFF
--- a/lib/rodauth/features/base.rb
+++ b/lib/rodauth/features/base.rb
@@ -712,7 +712,7 @@ module Rodauth
     # note that only the salt is returned.
     def get_password_hash
       if account_password_hash_column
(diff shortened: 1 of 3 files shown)


jeremyevans:
Thanks for the patch!  I agree that this makes things nicer.

PR #396, opened 2024-03-01
Update require login redirect when login route changes

janko:
When login feature is enabled, it's expected that requiring login will redirect to the login page. This won't be the case if `login_route` changes. I thought it would be convenient to have it automatically match, so that the developer doesn't have to also update `require_login_redirect` in that case.

### DIFF
--- a/lib/rodauth/features/base.rb
+++ b/lib/rodauth/features/base.rb
@@ -67,8 +67,6 @@ module Rodauth
     translatable_method :unverified_account_message, "unverified account, please verify account before logging in"
     auth_value_method :default_field_attributes, ''
 
-    redirect(:require_login){"#{prefix}/login"}
-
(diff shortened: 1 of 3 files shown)


jeremyevans (changes requested):
I'm fine with this change, but `redirect(:require_login){"#{prefix}/login"}` should be removed, and replaced with an appropriate `auth_value_method` argument.

janko:
I added the requested change.

jeremyevans:
My apologies, I didn't realize when I requested the change that the redirect call was in a different feature (base instead of login).  So your original patch did not redefine a method.  Your original patch was correct and I'll merge that.

jeremyevans:
Cherry-picked as 0869be21335c6892792922e878dd0b5b82e2fd33

PR #404, opened 2024-04-04
Skip default session clearing in JWT mode

janko:
Originally, `#clear_session` in JWT feature was calling `super` to clear the session hash from the instance variable. However, in https://github.com/jeremyevans/rodauth/pull/140 an explicit `session.clear` was added, because `super` didn't handle sessions plugin being loaded.

This means it's not necessary to call `super` anymore. This behavior caused a [bug](https://github.com/janko/rodauth-omniauth/discussions/13#discussioncomment-8961467) when `rodauth-omniauth` is used with `rodauth-rails`, where `super` being called when clearing JWT session called Rails implementation. The root cause was ultimately in `rodauth-omniauth`, but I thought it would be useful to have this change as a safeguard for any plugin extending `#clear_session`.

### DIFF
--- a/lib/rodauth/features/jwt.rb
+++ b/lib/rodauth/features/jwt.rb
@@ -59,11 +59,9 @@ module Rodauth
     end
 
     def clear_session
-      super
-      if use_jwt?
-        session.clear
-        set_jwt
-      end
+      return super unless use_jwt?
+      session.clear
+      set_jwt
     end
 
     def jwt_secret


jeremyevans:
Thanks for the patch!

Not calling `super` seems undesirable at first glance.  Overriding `clear_session` in the Rodauth config would be fine as that would be called before the `jwt` feature implementation, and no other feature seems to override the default `clear_session` in the base feature (except the internal_request feature, which always overrides without calling super).  Therefore, I don't think this should cause problems with core Rodauth, even if it seems initially undesirable.

Do you think that external Rodauth features that implement `clear_session` would expect their implementation to be ignored if `jwt` feature was used?  If so, at the very least, we should document this behavior in the jwt feature documentation.

PR #413, opened 2024-04-16
s/succesful/successful/g

jf:
Just noticed several typos for `succesful*`, so fixing these


jeremyevans:
Thanks for the patch!

jf:
:) thank you for the excellent software (not just Rodauth!) and support!

PR #414, opened 2024-04-19
prettify/simplify psql statements for `Using non-default schema` scenario

jf:
- simplify the `CREATE SCHEMA` statements (no need to repeat the schema name if `AUTHORIZATION` is the same
- prettify horizontal alignment (yeah sorry about the commit message; it's _horizontal_, not _vertical!_) so that it's easier (IMO) to see what's going on

(For the record, I did confirm with the pg docs ([15](https://www.postgresql.org/docs/15/sql-createschema.html) onwards) that this way of specifying `CREATE SCHEMA` works where you do not need to separately specify the schema name if the authorization is the same value (search for `Create a schema for user joe; the schema will also be named joe:`))


jeremyevans:

PR #418, opened 2024-04-26
lib/rodauth/features/base.rb: add `throw_status` method that throws :rodauth_error without needing to set_field_error

jf:
as per https://github.com/jeremyevans/rodauth/discussions/412

### DIFF
--- a/lib/rodauth/features/base.rb
+++ b/lib/rodauth/features/base.rb
@@ -648,6 +648,11 @@ module Rodauth
       throw :rodauth_error
     end
 
+    def throw_status(status)
+      set_response_error_status(status)
+      throw :rodauth_error
+    end
+
     def throw_error_status(status, field, error)
       set_response_error_status(status)
       throw_error(field, error)
--- a/lib/rodauth/features/base.rb
+++ b/lib/rodauth/features/base.rb
@@ -643,14 +643,13 @@ module Rodauth
       set_response_error_status(status)
     end
 
-    def throw_error(field, error)
-      set_field_error(field, error)
+    def throw_rodauth_error
       throw :rodauth_error
     end
 
-    def throw_status(status)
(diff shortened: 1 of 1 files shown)


jeremyevans:
The method I said I would accept in https://github.com/jeremyevans/rodauth/discussions/412#discussioncomment-9182587 would just call `throw :rodauth_error`.  Something like:

```ruby
def throw_error
  throw :rodauth_error
end
```

You could then write your own `throw_status` method that calls this method.  I would not accept a `throw_status` method, because there is nothing internal to Rodauth that uses it, as I mentioned in https://github.com/jeremyevans/rodauth/discussions/412#discussioncomment-9110491.

jf:
I see. Thanks for the clarification. I'll update the PR

jf:
hm, just realized that there is a `throw_error` method already in `lib/rodauth/features/base.rb` that has 2 arguments (`field`, and `error`). I could call the new method `throw_rodauth_error`... but I'm not sure I like this name / approach. What do you think of making `field` and `error` optional for `throw_error` like so?

```
def throw_error(field=nil, error=nil)
  set_field_error(field, error) unless field.nil? and error.nil?
end
```

jeremyevans:
I already made it clear in https://github.com/jeremyevans/rodauth/discussions/412#discussioncomment-9182587 that 
I don't want to make the `throw_error` arguments optional.  I would be OK adding `throw_rodauth_error` as a separate method.

jf:
My apologies. In my mind we were talking about modifying `throw_error_status` (https://github.com/jeremyevans/rodauth/discussions/412#discussioncomment-9172187); I thought things might be different for `throw_error`.

I've updated the PR to add `throw_rodauth_error`.

jeremyevans:
Squash merged at b44b441

PR #420, opened 2024-05-01
set nice autofocus on login/password field (autofocus will go to first field with attribute set)

jf:
Note:
- I did not opt for the same approach as for `autocomplete`, `inputmode`, etc. because for HTML, setting `autofocus="<any string>"` will only serve to activate `autofocus`
- because it is ok to have `autofocus` on multiple fields (only the first field will get `autofocus`-ed), I opted to have `autofocus` for both the login, and the password field, so when `rodauth.skip_login_field_on_login?` is true, the autofocus then goes to the password field

### DIFF
--- a/lib/rodauth/features/base.rb
+++ b/lib/rodauth/features/base.rb
@@ -206,7 +206,7 @@ module Rodauth
         required = "required=\"required\""
       end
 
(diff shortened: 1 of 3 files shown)


jf:
ok. I didnt think about the other forms, but yes, that makes sense. I will get to this soon-ish, but give me some time.

jeremyevans:
@jf are you still planning to work on this?

jeremyevans:
No problem, thanks for getting back to me.

PR #428, opened 2024-06-23
Fix typo - migation -> migration

mlpinit:
👋🏼 Noticed a typo while reading the README. Thanks for the great work with roadauth!


jeremyevans:
Thanks for the patch!

PR #431, opened 2024-07-04
Add `account_from_id` for loading account by id

janko:
This is useful in when delivering emails in a background job, where the background job is given the account ID. In that case, the Rodauth instance needs to be initialized outside of a request, and account and token loaded into it.

This is what rodauth-rails has been doing so far:

```rb
rodauth = RodauthApp.rodauth.allocate
rodauth.instance_eval { @account = account_ds(account_id).first! }
```

With the new method it can be cleaner:

```rb
rodauth = RodauthApp.rodauth.allocate
rodauth.account_from_id(account_id)
```

There are also a few places within Rodauth that can make use of this method.

### DIFF
--- a/doc/base.rdoc
+++ b/doc/base.rdoc
@@ -85,6 +85,7 @@ use_request_specific_csrf_tokens? :: Whether to use request-specific CSRF tokens
 
 == Auth Methods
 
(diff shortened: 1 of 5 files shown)


jeremyevans (commented):
Looks good to me.  I'll test and merge tomorrow.

PR #434, opened 2024-07-23
Allow sending webauthn modify emails outside of a request

janko:
In a background job, we need to allocate a Rodauth instance without any Rack request information if we wanted to use it to render emails. There the session will not be available, but the account can be loaded into the Rodauth instance.

```rb
# this is how rodauth-rails recommends sending emails in background job
rodauth = RodauthApp.rodauth.allocate
rodauth.account_from_id(account_id)
rodauth.webauthn_authenticator_added_email_body
```

In order for the webauthn modify emails to render successfully in this case, `#webauthn_account_id` needs to take the loaded account if available.

Not sure if it makes sense writing tests for this scenario, as I'm not sure how common this approach is outside of rodauth-rails (I don't know of an alternative).

### DIFF
--- a/lib/rodauth/features/webauthn.rb
+++ b/lib/rodauth/features/webauthn.rb
@@ -489,7 +489,7 @@ module Rodauth
     end
 
     def webauthn_account_id
-      session_value
+      account ? account_id : session_value
     end
 
     def webauthn_user_ids_ds
--- a/lib/rodauth/features/webauthn_login.rb
+++ b/lib/rodauth/features/webauthn_login.rb
@@ -89,9 +89,5 @@ module Rodauth
       end
       forms
     end
-
-    def webauthn_account_id
-      super || account_id
-    end
   end
 end
--- a/lib/rodauth/features/webauthn_verify_account.rb
+++ b/lib/rodauth/features/webauthn_verify_account.rb
@@ -38,9 +38,5 @@ module Rodauth
       @webauthn_credential = webauthn_setup_credential_from_form_submission
       add_webauthn_credential(@webauthn_credential)
     end
-
-    def webauthn_account_id
-      super || account_id
-    end
   end
 end


jeremyevans:
Looks good, thanks for the patch.  I'll try to get this merged tomorrow.

janko:
That's a good point that with `internal_request` no changes are needed, since it sets the session value.

I'm still going back-and-forth on whether it makes more sense to recommend using the `internal_request` feature for background emails, or if it's an overkill, since we wouldn't be using anything else other than `internal_request_eval`, and it would mean every rodauth-rails app would have to have that feature enabled (since you always want emails to be delivered in the background). 

I actually ended up using ORM associations to retrieve the number of WebAuthn credentials for the WebAuthn modify email templates in rodauth-rails ([commit](https://github.com/janko/rodauth-rails/commit/36191e0cce76ffd05af5829216f1ed49e1686c17)), so I didn't end up calling `rodauth.account_webauthn_ids`. Since this change brings security concerns, but in the end is not needed for rodauth-rails, I will close the PR 👍🏻

jeremyevans:
If this is only an issue with background emails, I think the best approach would be to have a Mail delivery_method that delivers in the background.  We could add the ability to set the delivery_method for emails sent by Rodauth. That way you would not need any changes to Rodauth to handle background emails.  The email would be built during the request, but still sent in the background.  What are your thoughts on that?

jeremyevans:
Maybe this is actually as simple as `send_email(&:deliver_later)`?

janko:
I could open a separate discussion thread for this. One issue with this approach is that backgrounding libraries usually log their job arguments, which means that the whole email would be logged by default. That wouldn't be secure, as that includes PII (emails), which is probably not GDPR-approved, but also actual email tokens that an attacker could find and use. Additionally, this would produce relatively large job arguments, which might increase storage usage on the job DB (Redis, SQL). Finally, this wouldn't solve cases where a 3rd-party service (SendGrid, Mailchimp) is used for generating & delivering emails, in which case Mail gem is skipped.

In Rails apps, emails are idiomatically handled using Action Mailer (which uses the Mail gem), which uses expected Rails view templating and background delivery via Active Job. rodauth-rails actually [uses an internal mailer](https://github.com/janko/rodauth-rails/blob/36191e0cce76ffd05af5829216f1ed49e1686c17/lib/rodauth/rails/feature/email.rb) for all emails by default, in the non-secure way described above. However, it also generates Rodauth configuration that overrides this with a custom mailer, where only account ID and raw token are sent to the mailer. This means when the `send_email(&:deliver_later)` you mentioned is configured, only the account ID and raw token are sent to the background job (and potentially logged).

```rb
create_reset_password_email do
  RodauthMailer.reset_password(account_id, reset_password_key_value)
end
```
```rb
class RodauthMailer < ActionMailer::Base
  def reset_password(account_id, key)
    rodauth = RodauthApp.rodauth.allocate
    rodauth.account_from_id(account_id)
    rodauth.instance_eval { @verify_account_key_value = key }

    mail subject: rodauth.reset_password_email_subject, to: rodauth.email_to, from: rodauth.email_from
  end
end
```

That's the safest way I found, but it does require overriding each email configuration method with the custom mailer implementation.

PR #444, opened 2024-10-11
Add rodauth-openapi to the website

janko:
It's not technically a "feature", since it doesn't provide any authentication functionality. It's more like tooling, and it made sense to put it together with the Postman documentation, since it's an alternative to that. While the Postman documentation is static and doesn't adapt to route/param changes, this is dynamic based on your actual Rodauth configuration.


jeremyevans:
Thank you very much for the patch, and for working on this!

PR #445, opened 2024-10-21
Allow disabling WebAuthn autofill UI

janko:
This is useful when the developer wants the ability to offer WebAuthn authentication without asking for login, but considers the autofill UI too magical for the end user, and would prefer to display an explicit "login via passkey" button instead.

The autofill UI has become even more magical in recent browser/OS versions, where the user is asked to authenticate via WebAuthn as soon as the login page is loaded. Previously, Chrome would display a passkey option in a dropdown after focusing the login field, but this is no longer the case.

This ties onto our discussion in https://github.com/jeremyevans/rodauth/discussions/363.

### DIFF
--- a/doc/webauthn_autofill.rdoc
+++ b/doc/webauthn_autofill.rdoc
@@ -10,6 +10,7 @@ independently from the autofill UI.
 
 == Auth Value Methods
 
+webauthn_autofill? :: Whether to activate the autofill UI on the login page.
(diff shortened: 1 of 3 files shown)


jeremyevans:
Thanks for the patch!  This looks good, and I agree that the feature makes sense.

I'll need to fix the CI gemfile for Ruby 2.3.

PR #449, opened 2024-11-12
Disable CSRF protection for JSON requests

janko:
Currently, CSRF protection is only disabled for JWT requests, as those use token-based instead of cookie-based authentication. However, JSON APIs that require `Content-Type: application/json` and have reasonable CORS restrictions don't appear to be vulnerable to CSRF attacks.

Since handling CSRF tokens isn't trivial to do on the client side, we can reduce friction by disabling CSRF protection by default for _all_ JSON requests (including JWT).

This is a follow-up to the discussion in https://github.com/jeremyevans/rodauth/discussions/448.

### DIFF
--- a/doc/json.rdoc
+++ b/doc/json.rdoc
@@ -23,11 +23,11 @@ to the JSON response:
     json_response[:error_reason] = reason
   end
 
-The session state is managed in the rack session, so make sure that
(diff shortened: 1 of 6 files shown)


jeremyevans:
This looks great, thank you for working on it.  I should have time to test and merge this later today.

PR #452, opened 2024-11-17
Add method for when 2nd factor auth is pending

janko:
This is useful for when applications want to modify configuration depending on whether 2FA is pending. For example, the Rails demo app currently redirects directly to 2FA after login:

```rb
login_redirect { uses_two_factor_authentication? && !two_factor_authenticated? ? two_factor_auth_required_redirect : super() }
login_notice_flash { uses_two_factor_authentication? && !two_factor_authenticated? ? "Please authenticate with 2nd factor" : super() }
```

(It's necessary to check `!two_factor_authenticated?` here as well, because passkey login can authenticate both factors at once.) With this method available, the code can be simplified:

```rb
login_redirect { two_factor_authentication_pending? ? two_factor_auth_required_redirect : super() }
login_notice_flash { two_factor_authentication_pending? ? "Please authenticate with 2nd factor" : super() }
```

### DIFF
--- a/demo-site/views/index.erb
+++ b/demo-site/views/index.erb
@@ -16,7 +16,7 @@
     <% if rodauth.logged_in_via_remember_key? %>
       <li><a href="/confirm-password">Confirm Password</a></li>
     <% end %>
-    <% if rodauth.uses_two_factor_authentication? && !rodauth.two_factor_authenticated? %>
+    <% if rodauth.two_factor_partially_authenticated? %>
       <li><a href="/multifactor-auth">Authenticate Using Additional Factor</a></li>
     <% else %>
       <li><a href="/multifactor-manage">Manage Multifactor Authentication</a></li>
--- a/lib/rodauth/features/two_factor_base.rb
+++ b/lib/rodauth/features/two_factor_base.rb
@@ -124,23 +124,12 @@ module Rodauth
     end
 
     def authenticated?
-      # False if not authenticated via single factor
-      return false unless super
-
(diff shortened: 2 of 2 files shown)


jeremyevans (commented):
I'm in favor of this idea, but I would like a better method name.  The issue I have with `pending?` is that Rodauth is designed to handle cases where certain actions do not require 2FA, and other cases do.  Just because 2FA is available does not mean that it is required by the application.

janko:
Yeah, I'm totally open to the naming change, as I was struggling with that precisely for the reason you mentioned. My previous ideas were `two_factor_authentication_required?` and `needs_two_factor_authentication?`, but those reinforced more the idea that 2FA is required when it's not.

If I take your words verbatim, maybe `two_factor_authentication_available?` would be an improvement? Still looking for something a bit more precise, as this sounds similar to `uses_two_factor_authentication?`.

janko:
Or I can flip it around and call it 'two_factor_authenticaton_satisfied?`. I like this option best so far.

jeremyevans:
`two_factor_authentication_satisfied?` seems weird to return true if two factor is not setup.

The best name I can think of  `two_factor_partially_authenticated?`, with semantics similar to the proposal of `two_factor_authentication_pending?`.  I think this makes sense, because if two factor is not setup, it would return false, and if it is two factor authenticated, it would also return false.    I think it would also have to check that you are partially authenticated (so that at least one authentication factor is present), which means it would also need a `require_login?` check.  What are your thoughts on that?

PR #454, opened 2024-11-18
Add guide for configuring case insensitive logins

janko:
As discussed [here](https://github.com/jeremyevans/rodauth/commit/eabfa697e9cf1871d6468c9940685579ad568b6d#commitcomment-149180108), I added a guide for configuring case insensitive logins, as an alternative to a new feature.


jeremyevans:
Thanks for the patch!  Cherry-picked at 9a5fd583d1746cbff454337a305a1d7d1700c9e7

PR #458, opened 2024-12-09
Raise error when attempting to render, but rendering was disabled

janko:
When configuring Rodauth plugin with `render: false`, it could happen that Rodauth would attempt to render a built-in view/email template:

* attempting to render a built-in email template is possible for JSON or internal requests
* external gem might incorrectly attempt to render a view template in JSON or internal requests
* in Rails apps, people would typically use Action View for rendering view/email templates, and they might miss some when disabling Tilt

Currently, Rodauth raises a rather cryptic:

```
NoMethodError: undefined method `parse_template_opts' for an instance of #<Class:0x000000011d9e3f40>
```

I initially started adding the error in rodauth-rails, but I thought it would be beneficial to have this in Rodauth itself. And for this it made sense to me to add a custom error class.

Technically, this doesn't cover the `#button` case, but those are only rendered within view templates, so the error should already be raised by `#_view` beforehand.

### DIFF
--- a/lib/rodauth.rb
+++ b/lib/rodauth.rb
@@ -3,6 +3,8 @@
 require 'securerandom'
 
 module Rodauth
+  class ConfigurationError < StandardError; end
+
   def self.lib(opts={}, &block) 
     require 'roda'
     c = Class.new(Roda)
--- a/lib/rodauth/features/base.rb
+++ b/lib/rodauth/features/base.rb
@@ -930,6 +930,10 @@ module Rodauth
     end
 
     def _view(meth, page)
+      unless scope.respond_to?(meth)
+        raise ConfigurationError, "attempted to render a built-in view/email template (#{page.inspect}), but rendering is disabled"
+      end
+
       scope.send(meth, _view_opts(page))
     end
   end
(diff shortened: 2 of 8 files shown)


jeremyevans (commented):
Thanks for the patch. Can we rename the custom error class from `Error` to `ConfigurationError` to make clear this is an error in the user's configuration?

janko:
Agreed, renamed 👍🏻 

I was thinking that afterwards we could reuse this error in other places:

https://github.com/jeremyevans/rodauth/blob/1b4eba51fe12587a044d90c24d038ceb880c13db/lib/rodauth/features/base.rb#L886-L889 https://github.com/jeremyevans/rodauth/blob/1b4eba51fe12587a044d90c24d038ceb880c13db/lib/rodauth/features/base.rb#L768-L771 https://github.com/jeremyevans/rodauth/blob/1b4eba51fe12587a044d90c24d038ceb880c13db/lib/rodauth/features/jwt.rb#L66-L69 https://github.com/jeremyevans/rodauth/blob/1b4eba51fe12587a044d90c24d038ceb880c13db/lib/rodauth/features/sms_codes.rb#L516-L518

jeremyevans:
All of those locations are great places to use this error, thank you for doing the research.

janko:
Updated the mentioned places with the new configuration error.

PR #460, opened 2024-12-10
Add guide on customizing email requirements

janko:
It's useful to know how to extend email requirements, for example verify whether the email exists to guard against spammers. It complements the existing guide on customizing password requirements.


jeremyevans:
Looks good, thank you very much!

PR #461, opened 2024-12-26
Make rodauth.has_password? method public

enescakir:
I needed this method when integrating rodauth-omniauth into our project.

If an account is created via omniauth, it doesn't have a password. In this case, I prefer to display "Create Password" instead of "Change Password".

Additionally, if the omniauth provider is the last authentication method, including the password, I aim to block to disconnect it.

Therefore, I believe this method is beneficial to have in public.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,5 @@
 === master
+* Make rodauth.has_password? method public (enescakir) (#461)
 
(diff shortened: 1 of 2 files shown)


jeremyevans:
Cherry-picked at bb37118a47a13dd74c2be0a370da398fdb554313

PR #462, opened 2024-12-26
Fix compatibility with JWT 2.10+

janko:
In JWT 2.10.0, the `JWT::VERSION::{MAJOR,MINOR,TINY,PRE}` constants were silently made private (see https://github.com/jwt/ruby-jwt/commit/e180765b56cb06407b39a6291d7364176f58dc0a), so use `JWT.gem_version` instead.

### DIFF
--- a/lib/rodauth/features/jwt.rb
+++ b/lib/rodauth/features/jwt.rb
@@ -105,7 +105,7 @@ module Rodauth
       jwt_decode_opts
     end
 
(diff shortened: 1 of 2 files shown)


jeremyevans:
Thanks for the patch! The jwt gem reversed the change in 2.10.1, so this isn't needed.  However, I checked and it seems all jwt versions support `JWT.gem_version`, so there shouldn't be harm in switching to that.

PR #471, opened 2025-02-23
docs: `configuration_eval` -> `configuration_module_eval`

bjeanes:
I was looking at the docs for custom features today and noted an example usage of `configuration_eval`. I wasn't totally sure how it should be used so searched the source for it and couldn't find any usages. Digging into the source code and the commit which introduced this term to the changelog and README, I think this was a typo for `configuration_module_eval`.

### DIFF
--- a/README.rdoc
+++ b/README.rdoc
@@ -1514,7 +1514,7 @@ use the following basic structure
         end
       end
 
-      configuration_eval do
+      configuration_module_eval do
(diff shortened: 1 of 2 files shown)


jeremyevans:
Thank you for the fix.  The method was probably named `configuration_eval` at some point during development, but changed to `configuration_module_eval` before it was committed.

PR #474, opened 2025-04-07
Fix typo in README

fkmy:
Just noticed several typos in the README, so fixing these.


jeremyevans:
Thanks for the patch!

PR #499, opened 2026-01-28
Optional Implicit Verification of accounts when successfully resetting password

manfrin:
In the common user flow of 

user signs up -> user forgets to ever verify -> user leaves for a while and comes back -> user has forgotten their password so tries resetting 

the system as works now will prevent them from resetting their password or ever logging in unless the implementing system is conscious of catching the error and routing the user through a verification flow. However, even in the case of the implementing system doing this, the flow is nightmare ux. The user will have to enter their email to verify, go to their email and click the link to verify, then once that's done, request a password reset, go back to their email once again, click the link, and then enter their new password. 

It's my belief that a user resetting their password is implicitly verifying their account email, as the very act of resetting a password *requires* the email to be real. 

This feature adds an optional and defaulted to `false` `reset_password_implicitly_verifies?` which handles this preferred user path. 

-----------------

I tried to keep it as simple and as bound to the style/convention of the codebase as possible. I did have a doubt as to whether this should be a toggle on :ResetPassword or its own feature definition (as it also depends on the verification feature which reset password does not automatically do). I opted for the former since it's basically a toggle check, but if you think it's more appropriate as a feature let me know and I can redo it.

### DIFF
--- a/lib/rodauth/features/reset_password.rb
+++ b/lib/rodauth/features/reset_password.rb
@@ -15,7 +15,7 @@ module Rodauth
     view 'reset-password-request', 'Request Password Reset', 'reset_password_request'
     additional_form_tags
     additional_form_tags 'reset_password_request'
-    before
+    before 
     before 'reset_password_request'
     after
     after 'reset_password_request'
@@ -133,7 +133,7 @@ module Rodauth
             throw_error_status(invalid_field_error_status, password_param, password_does_not_meet_requirements_message)
           end
 
-          if password_match?(password)
+          if password_match?(password) 
             throw_error_reason(:same_as_existing_password, invalid_field_error_status, password_param, same_as_existing_password_message)
           end
 
@@ -167,7 +167,7 @@ module Rodauth
           set_reset_password_email_last_sent
           @reset_password_key_value = reset_password_key_value
         elsif e = raised_uniqueness_violation{password_reset_ds.insert(reset_password_key_insert_hash)}
-          # If inserting into the reset password table causes a violation, we can pull the
+          # If inserting into the reset password table causes a violation, we can pull the 
           # existing reset password key from the table, or reraise.
           raise e unless @reset_password_key_value = get_password_reset_key(account_id)
         end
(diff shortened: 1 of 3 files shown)


jeremyevans:
Thank you for submitting a PR. I agree in principle that a valid password reset should implicitly verify. The whole point of account verification is to ensure the email is valid, and if the user can prove receipt of a password reset email, I think that's sufficient for account verification. I'll do a deeper review later today, and consider whether this should be a separate feature or just part of an existing feature.

lib/rodauth/features/reset_password.rb
```diff
@@ -231,6 +232,14 @@ def after_login_failure
       super
     end
 
+    def after_reset_password
+      super if defined?(super)
+      if reset_password_implicitly_verifies? && !open_account?
+        verify_account
+        remove_verify_account_key
```
jeremyevans, line 239: The main issue here is you can set `reset_password_implicitly_verifies?` to true without loading the `verify_account` feature, in which case you end up with a `NoMethodError` here. It's possible to check that, but I think it's simpler to add a new feature that depends on both `reset_password` and `verify_account`, and has these changes (as you mentioned when submitting).

lib/rodauth/features/reset_password.rb
```diff
@@ -254,7 +263,13 @@ def password_reset_ds(id=account_id)
     end
 
     def _account_from_reset_password_key(token)
-      account_from_key(token, account_open_status_value){|id| get_password_reset_key(id)}
+      account_from_key(token, reset_password_account_status_value){|id| get_password_reset_key(id)}
+    end
+
+    def reset_password_account_status_value
```
jeremyevans, line 269: You can keep this method inside the `reset_password` feature (and the call inside `_account_from_reset_password_key`) and have it return `account_open_status_value`, and extend it in the `reset_password_verifies_accounts` feature to do `[super, account_unverified_status_value].flatten`.

jeremyevans (commented):
This looks good, but after review, I think it's more appropriate as a separate feature. If you want to work on changing this PR to use a separate feature, that works for me. Otherwise, I'm happy to merge this PR and handle those changes.

jeremyevans:
I merged this and then moved the code to a new feature. Thank you for your contribution!

PR #502, opened 2026-04-19
Add after_no_matching_login hook

muellerj:
In order to log failed login attempts, a callback for a non-matching login would be helpful. 

Please advise if there is a better place to put this.


jeremyevans (commented):
Thank you for the patch. It seems reasonable to add this. However, could you please remove the unrelated whitespace changes, and add a spec?

muellerj:
Of course. I'll adjust the patch tomorrow.   
  
  
  
  
  
  
  

  
  
>

muellerj:
@jeremyevans: Please have a look at your convenience.

jeremyevans:
@muellerj I apologize for the long delay in getting to this. I was at RubyKaigi all last week. The changes look good, so this should get merged shortly. There are some CI errors related to rack that I'll work around.

PR #504, opened 2026-05-05
Fix `jwt_refresh` refresh endpoint returning 400 but also a token

tomasmiguez:
With our current setup:

```ruby
        class RodauthApp < Roda
          use Middlewares::AccountSelector

          plugin :rodauth, json: :only do
            enable :login, :jwt, :jwt_refresh, :session_expiration

            prefix '/api'
[...]
            allow_refresh_with_expired_jwt_access_token? true

[...]
          end

          route do |r|
            rodauth.check_session_expiration
            r.rodauth
          end
        end
```

which we mount on the main app under `/api`, we are getting the following behaviour:

```bash
curl -v http://localhost:3000/api/jwt-refresh -H 'authorization: Bearer validAccessToken' -H 'content-type: application/json' -d '{"refreshToken":"invalidRefreshToken"}'
[...]
< HTTP/1.1 401 Unauthorized
< authorization: newAccessToken
[...]
{"token":"newAccessToken","errors":[{"status":"401","code":"unauthorized","detail":"invalid JWT refresh token"}]}%
```

The `session_expiration` feature's `check_session_expiration` runs before `r.rodauth` and calls [`set_session_value(last_session_activity_at, now)`](https://github.com/jeremyevans/rodauth/blob/665751bf721fc023123f8630f632b55f9827500a/lib/rodauth/features/session_expiration.rb#L31) to bump activity. [`set_session_value`](https://github.com/jeremyevans/rodauth/blob/665751bf721fc023123f8630f632b55f9827500a/lib/rodauth/features/jwt.rb#L133-L137) chains into `set_jwt` → `set_jwt_token`, which [`jwt_refresh` overrides](https://github.com/jeremyevans/rodauth/blob/665751bf721fc023123f8630f632b55f9827500a/lib/rodauth/features/jwt_refresh.rb#L70-L77) to write the fresh JWT into both the authorization header and json_response['token'].
At that point `json_response[json_response_error_key]` isn't set yet, so the override takes the "no error" branch and stamps the body. The refresh route then enters the invalid-token branch and sets the error, but it calls [`_return_json_response`](https://github.com/jeremyevans/rodauth/blob/665751bf721fc023123f8630f632b55f9827500a/lib/rodauth/features/json.rb#L210-L214) which bypasses [the jwt.rb `return_json_response`](https://github.com/jeremyevans/rodauth/blob/665751bf721fc023123f8630f632b55f9827500a/lib/rodauth/features/jwt.rb#L145-L148) override that would re-run `set_jwt_token` and clean up the body.

The proposed fix makes sure the jwt is not set outside of Rodauth's routing, which solves the issue. The other alternative I can see is relying on `return_json_response` which would solve the issue as well, but would make redundant writes/checks.

### DIFF
--- a/lib/rodauth/features/jwt_refresh.rb
+++ b/lib/rodauth/features/jwt_refresh.rb
@@ -69,6 +69,9 @@ module Rodauth
 
     def set_jwt_token(token)
       super
+
+      return unless @current_route
+
       if json_response[json_response_error_key]
         json_response.delete(jwt_access_token_key)
       else
--- a/spec/jwt_refresh_spec.rb
+++ b/spec/jwt_refresh_spec.rb
@@ -680,4 +680,19 @@ describe 'Rodauth login feature' do
 
     json_request('/').must_equal [200, {"authenticated_by"=>["password"]}]
   end
+
+  it "should not leak access_token in the body when set_jwt_token is called outside a rodauth route" do
+    rodauth do
+      enable :login, :jwt_refresh, :session_expiration
+    end
+    roda(:jwt) do |r|
+      rodauth.check_session_expiration
+      r.rodauth
+    end
+
+    jwt_refresh_login
+
+    res = json_request("/jwt-refresh", :refresh_token=>'invalid')
+    res.must_equal [400, {"error"=>"invalid JWT refresh token"}]
+  end
 end


jeremyevans:
Thank you for submitting an issue. Can you explain why returning a new access token in this case is causing a problem?

tomasmiguez:
We like to keep payloads lean and clean on our system, and the token is not used in this case.
Also for other clients consuming our API, getting a token in the response of a refresh request that hasn't been refreshed could lead to ereoneous assumptions which we would like to minimize if possible.
Lastly, and without knowing the full intent behind the original implementation so I might be making invalid assumptions, it seems that the intent was for the jwt to not be set if there is an error, which only happens now due to an apparent optimizarion made by calling `_return_json_response` instead of the public version.

jeremyevans:
I looked at your example spec. Here's a related backtrace:

```
/home/jeremy/code/rodauth/lib/rodauth/features/jwt.rb:151:in 'Rodauth::Jwt#set_jwt'
/home/jeremy/code/rodauth/lib/rodauth/features/jwt.rb:135:in 'Rodauth::Jwt#set_session_value'
/home/jeremy/code/rodauth/lib/rodauth/features/session_expiration.rb:31:in 'Rodauth::SessionExpiration#check_session_expiration'                                                                                                        /home/jeremy/code/rodauth/spec/jwt_refresh_spec.rb:689:in 'block (3 levels) in <top (required)>'
```

The `session_expiration` feature is trying to do:

```ruby
set_session_value(session_last_activity_session_key, time)
```

What the pull request does in this case is to have the access token set in the response header (since you call `super`), but not in the response body. It seems like an odd choice to do that.

For `check_session_expiration` to work, it needs to update the session (the access token). Preventing this update from working will break things in at least some applications.

What happens here is that the access token is set successfully (updating the response headers and response body) when `check_session_expiration` is called. Later, the invalid jwt refresh error is used:

```
json_response[json_response_error_key] = jwt_refresh_invalid_token_message
```

If this happens, it sounds like you want to remove the access token from response header and response body?  Maybe we could add an `after_invalid_refresh_token` token that would allow you to do that?  I don't think it's something that Rodauth should do by default. The refresh token failure and the setting of the last activity time are orthogonal.

tomasmiguez:
Thanks for taking the time to review this issue! 

We currently solved this by redefining on the app this method:

```ruby
            set_jwt_token do |token|
              if json_response[json_response_error_key]
                json_response.delete(jwt_access_token_key)
              elsif @current_route
                set_response_header('authorization', token)
                json_response[jwt_access_token_key] = token
              end
            end
```

I brought it here to an issue to check if changing this behaviour was an improvement in general for other users, but it make sense that it might break some preexisting applications.

Thanks again!

jeremyevans:
Sounds good, I'm glad you have a fix that works for your app.

PR #505, opened 2026-06-01
Require `rack` in base feature so Rack.release is defined

fkmy:
## Problem
Loading the base feature raises `NoMethodError: undefined method 'release' for module Rack` when no other code in the process has done a full `require 'rack'` first.

[lib/rodauth/features/base.rb:589](https://github.com/jeremyevans/rodauth/blob/407dd7b191cf0028626136524c555e62f6416756/lib/rodauth/features/base.rb#L588) references `Rack.release`, but the file only requires `rack/request` and `rack/utils`. Since Rack 3.0, `Rack.release` is defined in `rack/version.rb`, and that file is loaded only by `lib/rack.rb` — no `rack/<subpath>` require pulls it in.

The bug is hidden in rodauth's own specs because `spec/spec_helper.rb` loads `rack/test` / `capybara`, which transitively `require 'rack'`. Clients that call `Rodauth::Auth.configure` without those (e.g. [rodauth-model](https://github.com/janko/rodauth-model)) hit it immediately.

### Reproduction
```ruby
require 'rodauth'
Class.new(Rodauth::Auth).configure { enable :base }
# => NoMethodError: undefined method 'release' for module Rack
```

## Fix
Replace the partial requires with `require 'rack'`. `Rack::Request` and `Rack::Utils` remain available since `lib/rack.rb` loads them.

### DIFF
--- a/lib/rodauth/features/base.rb
+++ b/lib/rodauth/features/base.rb
@@ -1,7 +1,17 @@
 # frozen-string-literal: true
 
-require 'rack/request'
-require 'rack/utils'
+begin
+  require "rack/version"
+rescue LoadError
+  require "rack"
+else
+  if Rack.release >= '3'
+    require "rack/request"
+    require "rack/utils"
+  else
+    require "rack"
+  end
+end
 
 module Rodauth
   Feature.define(:base, :Base) do


jeremyevans:
Thank you for the patch. However, the correct behavior when using Rack 3+ would be to require only the necessary parts of Rack, as requiring `rack` itself brings in autoloading that would be best to avoid. To be backwards compatible with older Rack versions, you need something like: https://github.com/jeremyevans/roda/blob/master/lib/roda/request.rb#L3-L13

fkmy:
@jeremyevans 
Thanks for the review! Updated to use a conditional require. https://github.com/jeremyevans/rodauth/commit/9cb5bf05c9560c9df7dee88ca56a7dbd2b0bb3e0

PR #509, opened 2026-07-12
Exclude nil instance variables from `#inspect` output on Ruby 4+

janko:
Now that all possible instance variables are initialized to `nil`, this inflates the `#inspect` output quite a bit, making it more difficult to focus on important details.

I think we can exclude those instance variables from the inspect output for brevity.


jeremyevans (approved):
Makes sense, thank you for the patch!

